PR #2, opened 2016-10-16
Greatly expand the "tutorial" documentation

timholy:


Evizero:
The new image `docs/src/assets/indexing/cmrot.png` does not seem to be used anywhere

docs/src/indexing.md
```diff
@@ -0,0 +1,123 @@
...
+
+julia> imgov = colorview(RGB, img0, imgrot, zeroarray);
+```
+
+![ref_overlay](assets/indexing/ref_overlay.png)
+
+The fact that the overlapping portion looks yellow---the combination
+of red and green---indicates that we have perfect alignment.
+
+You can learn more about Julia's support for arbitrary indices at ??. (**to be written**)
```
timholy: Glad it was useful. That's really intended for developers; I also plan a blog post on "cool stuff you can do with non-1 indices," and that's what I was really thinking about here.

timholy:
It was added before the `cmpair`; I'd rather not need `cmpair`, but I can't figure out how to get both images to display on the same line otherwise. CC @MichaelHatherly

Evizero:
One lazy way to do it (which is precisely what I did) would be to use tables. I do think they require headers, though, but I have never investigated.

Would look somewhat like this:
http://augmentorjl.readthedocs.io/en/latest/tutorials/augmentation.html#example-mnist-handwritten-digits

although I used rst, but markdown supports tables as well

MichaelHatherly:
Probably due to the `max-width: 48em` we set on `article`. @mortenpi, shall we increase the max width or just allow images to overflow if necessary? Not sure which will look nicer.

mortenpi:
> shall we increase the max width or just allow images to overflow if necessary?

It would be nice if the text wouldn't get wider. But I should try some some to figure out how to do this well.

timholy:
@MichaelHatherly I'd be happy if they dynamically shrink to fit.

@Evizero The table route seems OK, too. It seems that a column header is required, though (testing suggests I can't just use blank spaces).

MichaelHatherly:
> dynamically shrink to fit.

Not sure whether it's possible to do in a general way for an arbitrary numbers of images inline without resorting to js, if I'm understanding what you're wanting correctly. @mortenpi are you going to take a look at this one? I can add it to my list if not.

> It seems that a column header is required

Yes, I believe column headers are still required for our tables at the moment. Probably possible to remove that restriction, though you can get around it by using a non-breaking space, `Char(160)`, in the header to trick it into doing what you need.

PR #6, opened 2017-01-25
add juliaimages logo to docs

tlnagy:


timholy:
Good idea. This gets incorporated automatically?

tlnagy:
Yeah, Documenter.jl looks in `src/assets` for a file called `logo.png` to use. When I build it locally, it looks like this:

image

Issue #4, opened 2017-01-17
stable badge in readme


waldyrious:
Why was the "stable" badge removed from the readme in d22ffcc? It shows up in http://juliaimages.github.io along with the "latest" badge.

By the way, the link above should probably be added to the repository description.

timholy:
At the time, the stable link gave a 404 error. Now, it's been long enough since I read about this stuff that I don't even remember the difference between the two. One thing to keep in mind is that these docs aren't tied to a single package, they're for the whole ecosystem of packages, each of which can have a separate release schedule.

waldyrious:
Fixed in #13

PR #13, opened 2017-04-14
replace README.md --> index.html with a redirect

waldyrious:
As described in #4, visiting http://juliaimages.github.io/ rather than http://juliaimages.github.io/latest shows the contents of the README.md from the `master` branch.

Since on GitHub the default branch is already configured to be `source`, there's no need for a README in the `master` branch (the fact that it contains broken links to the docs further confirms that it was indeed not used as a README).

With this change, visiting http://juliaimages.github.io/ will automatically take people to http://juliaimages.github.io/latest rather than show them the rendered README page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,0 @@
-# juliaimages.github.io
-
-Documentation For JuliaImages
-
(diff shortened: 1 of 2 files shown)


timholy:
Seems like a fine idea to me. Any objections?

timholy:
Thanks!

PR #17, opened 2017-08-16
[WIP] Docs for ImageSegmentation.jl

annimesh2809:
Docs and tutorials for `ImageSegmentation.jl`.


annimesh2809:
@tejus-gupta Let's complete the docs and tutorials.

timholy:
Since time is running short (we're supposed to submit our final evaluation by Monday), do you think it would be possible to have a *complete* first draft by Saturday morning? Presumably I'll have comments on the first version :smile:, and only after those are addressed does it seem reasonable to copy/paste the material to build the blog post. I suspect the wider community of Julians may want to take a peek at the blog post before submitting a passing final evaluation. So even Saturday doesn't leave a lot of time.

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,205 @@
...
+
+#### Seeded Region Growing
+
+This algorithm segments an image with repsect to a set of *n* seeds. Given the
+set of seeds in the form of a vector of tuples of `CartesianIndex` and label,
+the algorithm tries to assign these labels to each of the remaining points.
+
+###### Inputs
+
+* `img`             :  N-D image to be segmented (arbitrary indices are allowed)
```
timholy: This kind of detailed information about the arguments could go in `function_reference.md`, and be taken directly from the docstring. I'd rather make this section more expository.

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,205 @@
...
+
+**Original:**
+
+![Original](assets/segmentation/camera.jpg)
+
+**Segmented Image:**
+
+![SegmentedImage](assets/segmentation/camera_seg.jpg)
+
+#### Mean Shift
```
timholy: Duplicated section

annimesh2809:
@timholy I think it is ready for a review.

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+| 0.02 | ![seg1](https://user-images.githubusercontent.com/15063205/27088052-41996c74-5074-11e7-9ff1-1941d1be88c4.jpg) | 50% |
+| 0.05 | ![seg2](https://user-images.githubusercontent.com/15063205/27088179-9761458c-5074-11e7-9305-91411d862b22.jpg) | 62.5% |
+| 0.1 | ![seg3](https://user-images.githubusercontent.com/15063205/27088318-f556be88-5074-11e7-9b1e-271963060e90.jpg) | 70.8% |
+| 0.2 | ![seg4](https://user-images.githubusercontent.com/15063205/27088449-4ed7774a-5075-11e7-91bd-f18438ca57a0.jpg) | 79.2% |
+
+
+#### Felzenswalb's Region Merging Algorithm
+
+This algorithm operates on a Region Adjacency Graph (RAG). Each pixel/region is a node in the graph and adjacent pixels/regions have edges between them with weight measuring the dissimilarity between pixels/regions. The algorithm repeatedly merges similar regions till we get the final segmentation. It efficiently computes oversegmented “superpixels” in an image. The function can be directly called with an image (the implementation internally creates a RAG of the image first and then proceeds).
```
timholy: Someday would be cool to have tools to visualize the RAG like http://melvincabatuan.github.io/RAG/. (Not a priority now...)

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+
+img = Gray.(testimage("house"));
+segments = felzenszwalb(img, 300, 100);
+```
+
+![img1](assets/segmentation/house.jpg) ![img2](assets/segmentation/felzenszwalb.jpg)
+
+#### MeanShift Segmentation
+
+MeanShift is a clustering technique. It's primary advantages are that it doesn't assume a prior on the shape of the cluster (e.g, gaussian for k-means) and we also don't need to know the number of clusters beforehand. The algorithm doesn't scale well with size of image.
```
timholy: It's->Its. If this doesn't get merged before the end of GSoC, we should probably omit this section anyway.

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+
+#### Fast Scanning
+
+Fast scanning algorithm tries to segment the image in two pass by comparing
+each pixel to its left-neighbour and noting if it can be merged with them. If it
```
timholy: them->it. "can be merged" might need a bit of amplification wrt the parameter.

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+can't be merged then a new label is assigned to it. If more than one labels can be
```
timholy: labels->label

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+
+```julia
+using Images, ImageSegmentation, TestImages;
+
+img = Gray.(testimage("house"));
+segments = felzenszwalb(img, 300, 100);
+```
+
+![img1](assets/segmentation/house.jpg) ![img2](assets/segmentation/felzenszwalb.jpg)
+
```
timholy: Not clear how the segmented image was visualized. It's possible that you should create some visualization functions at the top of this section, e.g., "To visualize each segment in terms of its mean color or intensity, ..." and "To assign a separate color to each segment, ..." and then just call these functions for each visualization.

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+#### Creating a Region Adjacency Graph (RAG)
+
+A region adjacency graph can directly be constructed from a `SegmentedImage`
+using the `region_adjacency_graph` function. Each segment is denoted by a vertex
+and edges are constructed between adjacent segments. The output is a tuple of
+`SimpleWeightedGraph` and a Dict(label=>vertex) with weights assigned according to `weight_fn`.
+
+```julia
+julia> using ImageSegmentation, Distances
+julia> weight_fn(i,j) = euclidean(seg.segment_pixel_count[i], seg.segment_pixel_count[j]);
```
timholy: In this section, it might be worth justifying why/whether the pixel count is an interesting/useful means for making decisions about merging. It's also implicit here that `seg` exists, but where did it come from?

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+julia> G, vert_map = region_adjacency_graph(seg, weight_fn);    # `seg` is a `SegmentedImage`
+julia> G
+{2536, 6536} undirected simple Int64 graph with Float64 weights
```
timholy: Can we visualize the graph? (Not worth spending much time on.)

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+`RegionTree`.
+
+```julia
+julia> using ImageSegmentation
+julia> function homogeneous(img)
+           min, max = extrema(img)
+           max - min < 0.2
+       end
+julia> t = region_tree(img, homogeneous)        # `img` is an image
+Cell: RegionTrees.HyperRectangle{2,Float64}([1.0, 1.0], [300.0, 300.0])
```
timholy: Cross-link to RegionTrees.jl?

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+
+```julia
+julia> seg.image_indexmap
+4×4 Array{Int64,2}:
+ 1  1  3  3
+ 1  1  3  3
+ 2  2  2  2
+ 2  2  2  2
+julia> diff_fn(rem_label, neigh_label) = seg.segment_pixel_count[rem_label] - seg.segment_pixel_count[neigh_label];
+julia> new_seg = prune_segments(seg, [3], diff_fn);
```
timholy: How does one get the whole `seg`? (E.g., if a user wants to try this.)

docs/src/imagesegmentation.md
tejus-gupta: wow, those visualizations were really good.

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+## Example
+
+Let's see an example on how to use the segmentation algorithms in this package. We will try to seperate the deer, the ground and the sky in the image below using seeded region growing algorithm. This algorithm initializes the regions as the user-provided seed points and iteratively adds pixels to each region.
+
+![Original](assets/segmentation/deer.jpg) 
+
+```julia
+using Images, ImageSegmentation, ImageView
+
+img = load("deer.jpg")
```
annimesh2809: I have changed the images that did not have the "resue with modification for non-commercial purposes" license. 

Should we provide a link to the source of the images?

timholy: Yes, a link is needed to ensure reproducibility.

Can one see a license statement somewhere for the new images? It doesn't happen often, but occasionally one gets asked to prove that you have rights to the image, and if it doesn't come with a license statement (or site policy) then that's not possible. (All the images at TestImages were selected because we could find a statement somewhere.)

annimesh2809: For all of the images (except worm.jpg), the sites explicitly mention that the images are free to use for commercial purposes under the CC0 Public Domain. worm.jpg is free to use under the Creative Commons Attribution-Share Alike 4.0 International license, [here](https://commons.wikimedia.org/wiki/File:Davidraju_Worm_Snake.jpg) is the license link.

waldyrious: @annimesh2809 I would strongly encourage you to link to the Wikimedia Commons page (https://commons.wikimedia.org/wiki/File:Davidraju_Worm_Snake.jpg) instead of imgbb.com (https://ibb.co/hR2aWk) for the source of the image. That ensures the licensing information is preserved here, as required by the CC-BY-SA license (e.g as a link to the page with further information) and generally is more useful to the end user, since the image there is properly categorized, identified, etc.

In fact I'd suggest replacing all the example images by ones from Wikimedia Commons, because it's one of the few image hosting websites that make the metadata, authorship and licensing information for images very visible and convenient to access, which might come handy in cases like the one @timholy mentioned ("occasionally one gets asked to prove that you have rights to the image", etc.)

All of the images there are freely usable under open licenses, and specific images can be found easily because they tend to be well categorized. I'd be happy to help you if you need assistance locating images for any particular subject.

timholy:
Just let me know when you want another round of review here. (There appear to be a number of unaddressed comments so I'll hold off until requested.)

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,317 @@
...
+
+![Original](assets/segmentation/deer_seg1.jpg) 
+
+All the segmentation algorithm (except Fuzzy C-means) return a struct `SegmentedImage` that stores the segmentation result. `SegmentedImage` contains a list of applied labels and an array containing the assigned label for each pixel. `seeded_region_growing` needs the initial seed points from the user. Some algorithms only need to know the number of segments or level of detail in segmentation. For example, `felzenszwalb` only needs a single parameter k which controls the size of segments. Larger k will result in bigger segments.
+
+```julia
+using Images, ImageSegmentation, ImageView
+
+img = load("deer.jpg")
+segments = felzenszwalb(img, 100)
```
tejus-gupta: ```min_size``` is the minimum segment size in pixels. 

From the paper
```
 In practice k sets a scale of observation, in that a larger k causes a preference
 for larger components. Note, however, that k is not a minimum component size. 
Smaller components are allowed when there is a sufficiently large difference between
 neighboring components.
```
I don't think ```k``` has any unit that is physically meaningful. I agree that we should add some information about reasonable values of ```k```.

docs/src/imagesegmentation.md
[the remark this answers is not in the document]
annimesh2809: @waldyrious The problem with linking to Wikimedia Commons page is the size of the image. I require a smaller version of that image so that it looks good on the tutorials page. For the `seeded_region_growing`, I need to mention some hard- coded co-ordinates of the image as seeds so that the user can try it if he wants. Is it possible to use this image from imgbb.com and add a reference to the original source of the image (Wikimedia) as a footnote?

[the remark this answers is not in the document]
waldyrious: I'll reply in the main PR thread for convenience, since this discussion is happening on an outdated code block that github collapses by default.

waldyrious:
(continuing an [inline discussion](https://github.com/JuliaImages/juliaimages.github.io/pull/17#discussion_r134109649) in the main PR thread, since the code is outdated so the discussion is collapsed by default)

> @waldyrious The problem with linking to Wikimedia Commons page is the size of the image. I require a smaller version of that image so that it looks good on the tutorials page. For the seeded_region_growing, I need to mention some hard- coded co-ordinates of the image as seeds so that the user can try it if he wants. Is it possible to use this image from imgbb.com and add a reference to the original source of the image (Wikimedia) as a footnote?

You can actually directly use a specific image size from Wikimedia Commons. If you look just below the image at https://commons.wikimedia.org/wiki/File:Davidraju_Worm_Snake.jpg, you'll see the following:

> Size of this preview: 800 × 533 pixels. Other resolutions: 320 × 213 pixels | 640 × 427 pixels | 1,024 × 683 pixels | 1,280 × 853 pixels | 5,184 × 3,456 pixels.

Clicking any of those will give you a resized image in those dimensions; in fact, by changing the pixel width in the URLs of those links, you can generate a version of the image at *any* size you wish, which then gets cached in the Wikimedia servers. Here's an example for a 100px-wide thumbnail: https://upload.wikimedia.org/wikipedia/commons/thumb/6/6d/Davidraju_Worm_Snake.jpg/100px-Davidraju_Worm_Snake.jpg

Hope that helps!

annimesh2809:
@waldyrious Awesome! This is what I wanted :smile:

waldyrious:
Glad to hear that :) are you planning to replace the other images as well? I can assist you in locating replacements if that's an issue.

annimesh2809:
@waldyrious We can replace the other images in some later PR. Currently the priority is to complete these docs as soon as possible (due to the GSoC deadline). Thanks for the help. I will ping you when we have to update the images.

annimesh2809:
@timholy I think we are ready for round 2!

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,459 @@
...
+**Original** [(source)](https://upload.wikimedia.org/wikipedia/commons/thumb/d/d9/Flower-631765_960_720.jpg/800px-Flower-631765_960_720.jpg)
+
+
+![Original](assets/segmentation/flower.jpg)
+
+**Output with pixel intensity = cluster center intensity * membership of pixel in that class**
+
+![SegmentedImage1](assets/segmentation/flower_s1.jpg) ![SegmentedImage2](assets/segmentation/flower_s2.jpg)
+![SegmentedImage3](assets/segmentation/flower_s3.jpg)
+
```
timholy: A few verbal comments might be in order. What do the 3 clusters "mean"? (Obvious to you, but worth explaining that cluster 1 corresponds to the magenta petals, cluster 2 to the leaves, and cluster 3 to the white background.) Also, out of curiosity why doesn't the "leaves cluster" look green?

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,459 @@
...
+using Images, ImageSegmentation;
+
+img = load(download("http://docs.opencv.org/3.1.0/water_coins.jpg"));
+bw = Gray.(img).>0.5;
+dist = 1.-distance_transform(feature_transform(bw));
+markers = label_components(dist.<-15);
+segments = watershed(dist, markers);
+```
+
+![img1](assets/segmentation/water_coins.jpg) ![img2](assets/segmentation/watershed.jpg)
```
timholy: Maybe display `dist` too? And put in a table with column headers?

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,490 @@
...
+
+## Installation
+
+```julia
+Pkg.add("ImageSegmentation")
+```
+
+## Example
+
+Image segmentation is not a mathematically well-defined problem: for example, the only lossless representation of the input image would be to say that each pixel is its own segment. Yet this does not correspond to our own intuitive notion that some pixels are naturally grouped together. As a consequence, many algorithms require parameters, often some kind of threshold expressing your willingness to tolerate a certain amount of variation among the pixels within a single segment.
```
timholy: I vacillate between keeping this here and moving it to the Introduction (after the first paragraph). Whatever you two decide, I'm fine with.

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,490 @@
...
+The documentation for `seeded_region_growing` says that it needs two arguments - the image to be segmented and a set of seed points for each region. The seed points have to be stored as a vector of `(position, label)` tuples, where `position` is a [`CartesianIndex`](https://docs.julialang.org/en/stable/manual/arrays/#Cartesian-indices-1) and `label` is an integer. We will start by opening the image using ImageView and reading the coordinates of the seed points.
+
+```julia
+using Images, ImageView
+
+img = load("horse.jpg")
+imshow(img)
+```
+
+Hover over the different objects you'd like to segment, and read out the coordinates of one or more points inside each object. We will store the seed points a vector of `(seed position, label)` tuples and use `seeded_region_growing` with the recorded seed points.
```
timholy: "*as* a vector"

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,490 @@
...
+
+```julia
+using ImageSegmentation
+seeds = [(CartesianIndex(126,81),1), (CartesianIndex(93,255),2), (CartesianIndex(213,97),3)]
+segments = seeded_region_growing(img, seeds)
+```
+
+All the segmentation algorithm (except Fuzzy C-means) return a struct `SegmentedImage` that stores the segmentation result. `SegmentedImage` contains a list of applied labels, an array containing the assigned label for each pixel and mean color and number of pixels in each segment. [This section](##Result) explains how to access information about the segments.
```
timholy: "All the segmentation algorithm*s* ..." (pluralize)

Also, add comma: "...for each pixel, and mean color and number of pixels..."

The Result link doesn't quite work. (You're presumably aware you can test via `include("make.jl")` in the top directory and then examining `build/index.html` in your browser.)

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,490 @@
...
+#first segment's color (horse) = RGB(0.0647831,0.0588508,0.074473) = black
+#second segment's color (sky) = RGB(0.793598,0.839543,0.932374) = light blue
+#third segment's color (grass) = RGB(0.329876,0.357805,0.23745) = green
+
+# for visualizing the segmentation, create an image by replacing each each label in label_map(segments) with it's mean color
+imshow(map(i->segment_means(segments,i), labels_map(segments)))
+```
+![Original](assets/segmentation/horse_seg1.jpg)
+
+You can see that the algorithm did a fairly good job of segmenting the three objects. The only obvious error is the fact that elements of the sky that were "framed" by the horse ended up being grouped with the ground. This is because seeded_region_growing always returns connected regions, and there is no path connecting those portions of sky to the larger image. If we add some additional seed points in those regions, and give them the same label 2 that we used for the rest of the sky, we will get a result that is more or less perfect.
```
timholy: Add backticks around `seeded_region_growing`

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,490 @@
...
+
+segments = felzenszwalb(img, 10)  #smaller segments but noisy segmentation
+imshow(map(i->segment_means(segments,i), labels_map(segments)))
+```
+
+| k = 100 | k = 10 |
+|:------:|:---:|
+| ![Original](assets/segmentation/horse_seg2.jpg) | ![Original](assets/segmentation/horse_seg3.jpg) |
+
+We only got two segments with k = 100. Setting k = 10 resulted in smaller but rather noisy segments. `felzenzwalb` also takes an optional argument `min_size` - it removes all segments smaller with less than `min_size` pixels. Most methods don't remove small segments in their core algorithm. We can use the `prune_segments` method to postprocess the segmentation result and remove small segments.
```
timholy: "smaller ~~with less~~ than"

timholy: Maybe put the last two sentences in parentheses? (optional, your call)

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,490 @@
...
+
+Let's see an example on how to use watershed to segment touching objects. To use watershed, we need to modify the image such that in the new image flooding the topographic surface from the markers separates each coin. If this modified image is noisy, flooding from local minima may lead to oversegmentation and so we also need a way to find the marker positions. In this example, the inverted distance transform of the thresholded image (dist image) has the required topographic structure ([This page](https://in.mathworks.com/company/newsletters/articles/the-watershed-transform-strategies-for-image-segmentation.html) explains why this works). We can threshold the 'dist' image to get the marker positions.
```
timholy: Add/fix backticks around `dist` (2 cases). Maybe change to a \[\`distance_transform\`\](`@ref`) link? (This will automatically link to the function reference section.)

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,490 @@
...
+
+###### Demo
+
+```julia
+using Images, ImageSegmentation;
+
+img = load(download("http://docs.opencv.org/3.1.0/water_coins.jpg"));
+bw = Gray.(img).>0.5;
+dist = 1.-distance_transform(feature_transform(bw));
```
timholy: Rather than doing `1 - ` and then thresholding as `.<-15` would it be more straightforward to simply say
```julia
dist = distance_transform(feature_transform(bw));
markers = label_components(dist .> 16);
```
? (This would of course change the images.)

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,490 @@
...
+
+
+## Some helpful functions
+
+#### Creating a Region Adjacency Graph (RAG)
+
+A region adjacency graph can directly be constructed from a `SegmentedImage`
+using the `region_adjacency_graph` function. Each segment is denoted by a vertex
+and edges are constructed between adjacent segments. The output is a tuple of
+`SimpleWeightedGraph` and a Dict(label=>vertex) with weights assigned according to `weight_fn`.
```
timholy: Add backticks around `Dict(label=>vertex)`

docs/src/imagesegmentation.md
```diff
@@ -0,0 +1,490 @@
...
+```
+
+For more information regarding `RegionTrees`, see [this](https://github.com/rdeits/RegionTrees.jl#regiontreesjl-quadtrees-octrees-and-their-n-dimensional-cousins).
+
+#### Pruning unnecessary segments
+
+All the unnecessary segments can be easily removed from a `SegmentedImage` using
+`prune_segments`. It removes a segment by replacing it with the neighbor which has
+the least value of `diff_fn`. A list of the segments to be removed can be supplied.
+Alternately, a function can be supplied that return true for the labels that must
```
timholy: return->returns. Maybe add backticks around `true`

docs/src/imagesegmentation.md
annimesh2809: The "leaves cluster" includes pixels other than green that are close to the cluster center in the RGB space. Also the "image pixel intensity = center intensity * membership" rather than just the center intensity, so color is kind of greenish-black.
greenish
In the above image, each pixel intensity denotes the membership of the pixel in the "leaves cluster". As can be seen, many pixels that are not green also contribute fairly to this cluster.

timholy:
The doc hosting may fail if ImageSegmentation isn't registered (unless we put a `Pkg.clone` in `.travis.yml`). Are we ready to register it?

docs/src/imagesegmentation.md
tejus-gupta: We get the correct ```markers``` this way but ```dist``` wouldn't be correct. (We are passing both ```dist``` and ```markers``` to watershed.)

docs/src/imagesegmentation.md
tejus-gupta: I think we should keep it here.

timholy:
Ready for squash and merge?

annimesh2809:
Yes from my side!

annimesh2809:
@timholy How do we prepare the blog post? Where should we place it?

timholy:
Yay, great work folks!

For the official julia blog, clone https://github.com/JuliaLang/julialang.github.com. Add the text to `blog/_posts`. Add any images to a new folder in `images/blog`. Unfortunately I don't know of a build script, but if you discover one let me know!

PR #18, opened 2017-08-26
Add documentation for ImageFeatures

timholy:
This extracts the BRISK example from the ImageFeatures documentation and copies it here, plus a bit of introductory text. It also adds ImageFeatures to the function reference.


Evizero:
Is this good to merge?

timholy:
Thanks for the reminder.

PR #19, opened 2017-10-13
Documenter updates

mortenpi:
Following up on https://github.com/JuliaImages/juliaimages.github.io/commit/5523086c2d1948f905085148aa1959393092a585#commitcomment-24839336


timholy:
Thanks so much!

PR #21, opened 2017-10-23
segment_means -> segment_mean

mauro3:


timholy:
Thanks! I wonder why this didn't show up as an error when building the docs? @annimesh2809 and @tejus-gupta?

PR #29, opened 2018-10-04
Update to julia 1.0; various fixes; add a "landing page" for demos

### LINKED ISSUE #27: Provide a minimum working example(even 3x3) at the quickstart without Random.
Coming from someone new to the language it took me a long time and a lot of scrolling to understand what part encodes for the color, and what component handles the positioning. The rand isn't making it easy and reproducible.

timholy:
I believe this may be the last step to bring JuliaImages into the 1.0 era!

This updates the docs to use julia-1.0 (fixes #27), and in the mean time fixes a number of other issues.

Going forward, one of the things I most want is a collection of demos like the ones in [scikit-image](http://scikit-image.org/docs/stable/auto_examples/). This should be a community effort. I've tried to kick-start the process by providing some instructions and creating a (primitive) thumbnails page. Improvements either to the infrastructure, or by adding more demos, would be greatly appreciated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ language: julia
 os:
   - linux
 julia:
-  - 0.6
+  - 1.0
 notifications:
   email: false
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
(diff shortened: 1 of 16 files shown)


timholy:
Linking to discussion with @Evizero about table formats for the demo "landing page": https://github.com/JuliaDocs/Documenter.jl/issues/423#issuecomment-427041107

Evizero:
I sketched out a possible solution for the thumbnails. Anyone feel free to copy and paste.

rep: https://github.com/Evizero/EvisLittleCheatsheet.jl
preview: https://evizero.github.io/EvisLittleCheatsheet.jl/latest/

Sorry that I didn't take the time to contribute this here directly, but I have more or less near zero time right now for Julia stuff (for one or two weeks)

PR #31, opened 2018-10-12
Update conversions_views.md to include StackedView

Tokazama:
Add documentation as referenced [here](https://github.com/JuliaImages/Images.jl/issues/754)

### DIFF
--- a/docs/src/conversions_views.md
+++ b/docs/src/conversions_views.md
@@ -410,6 +410,29 @@ compare two (or more) grayscale images. See
 [Keeping track of location with unconventional indices](@ref) for more
 information.
 
+## StackedViews
+
+Sometimes it's helpful to combine several images into a single view for 
+further array like manipulation. 
+```jldoctest; setup = :(using Images)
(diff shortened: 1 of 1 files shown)


docs/src/conversions_views.md
```diff
@@ -410,6 +410,29 @@ compare two (or more) grayscale images. See
 [Keeping track of location with unconventional indices](@ref) for more
 information.
 
+## StackedViews
+
+Sometimes it's helpful to combine several images into a single view for 
+further array like manipulation. 
+```jldoctest; setup = :(using Images)
+julia> img1 = ImageMeta(rand(2,2,2))
```
timholy: How about something deterministic like `reshape(0.01:0.01:0.08, (2,2,2))`? That way it can pass the doctest. (Here and in the next line too.)

timholy:
Note if you're not a git wizard and find fixing it intimidating, let me know I can help or just do it for you.

timholy:
See #32. Would like to know if that revision is OK.

Tokazama:
That looks perferct

PR #32, opened 2018-10-15
Document StackedViews

timholy:
This is #31 with a couple of tweaks. I'm not quite sure why there were still `doctest` errors, but my (surprising) hunch it was the blank line after the ` ```jldoctest``` `. In the course of debugging I switched from ImageMeta to plain ranges, in case it had something to do with the more complex output of an ImageMeta. Since that type isn't introduced until later in the documentation anyway, that's probably good on its own.

@Tokazama, if you approve I will squash this and merge it giving you credit for the PR.

### DIFF
--- a/docs/src/conversions_views.md
+++ b/docs/src/conversions_views.md
@@ -412,26 +412,42 @@ information.
 
 ## StackedViews
 
-Sometimes it's helpful to combine several images into a single view for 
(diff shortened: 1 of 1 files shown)


timholy:
Merged manually in f38bf052894c06b98e0c220efce859709178a117

PR #38, opened 2018-12-26
Add missing function

aquatiko:
I found the implementation at: https://github.com/JuliaImages/Images.jl/blob/96e10ad90211222dbb95afe3e0f523b6139b8eca/src/algorithms.jl#L942


timholy:
Many thanks!

PR #41, opened 2019-01-06
Add mask step

aquatiko:
Fixes https://github.com/JuliaImages/ImageSegmentation.jl/issues/22

### DIFF
--- a/docs/src/imagesegmentation.md
+++ b/docs/src/imagesegmentation.md
@@ -452,7 +452,7 @@ Segmented Image with:
   labels map: 312×252 Array{Int64,2}
   number of labels: 24
 
-julia> segments = imshow( map(i->segment_mean(segments,i), labels_map(segments)) .* .~bw)       #shows segmented image
+julia> imshow( map(i->segment_mean(segments,i), labels_map(segments)) .* .~bw)       #shows segmented image
 ```
 
 | Original Image | Thresholded Image |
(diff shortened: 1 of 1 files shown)


docs/src/imagesegmentation.md
```diff
@@ -451,6 +451,8 @@ julia> segments = watershed(dist, markers)
 Segmented Image with:
   labels map: 312×252 Array{Int64,2}
   number of labels: 24
+
+julia> segments = imshow( map(i->segment_mean(segments,i), labels_map(segments)) .* .~bw)       #shows segmented image
```
Evizero: I don't know this code snippet, but the `segments = ...` prefix doesn't quite look right

aquatiko: @Evizero Updated. The code snippet is from the docs only, I have just added a mask to show the segmented image.

tejus-gupta:
@aquatiko Thanks for the PR.

The image in the docs was obtained by replacing each label with a random colour. Better to update the example code to this.

```
using ImageView, Random
function get_random_color(seed)
    Random.seed!(seed)
    rand(RGB{N0f8})
end

# remove labels from background region
imshow(map(i->get_random_color(i), labels_map(segments).*(1 .-bw)))
```

tejus-gupta:
LGTM!

PR #42, opened 2019-01-11
added demonstration: RGB to HSV and thresholding

aquatiko:
This demo shows how to convert RGB to HSV and threshold images based on their respective hue, saturation and value.


docs/src/demos/rgb_hsv_thresholding.md
```diff
@@ -0,0 +1,37 @@
...
+A simple segmentation of the image can then be effectively performed 
+by a mere thresholding of the HSV channels.
+
+
+```julia
+using Images, TestImages, LinearAlgebra, Interact, ImageView
+
+rbg_img = testimage("lighthouse")
+hsv_img = HSV.(rbg_img)
+channels = float(channelview(img))
```
Evizero: what is `img` ? should this be `hsv_img`?

docs/src/demos/rgb_hsv_thresholding.md
```diff
@@ -0,0 +1,37 @@
...
+
+hue_img = channels[1,:,:]
+value_img = channels[3,:,:]
+saturation_img = channels[2,:,:]
+@manipulate for hue in 0:255, saturation in 0:255, value in 0:255
+    binary_img = zeros(size(hue_img))
```
Evizero: can this be preallocated outside of `@manipulate` or does that break the code somehow?

aquatiko: @Evizero It won't work outside, the `binary_img` will change after each thresholding and the array needs to be reset each time.

Evizero: right. I was more talking about moving the allocation outside and using `fill!(binary_img, 0.0)` inside instead (which doesn't allocate)

aquatiko: @Evizero I wasn't quite aware the method. I pushed the suggested changes. 
Just a question out of curiosity,
Does this speeds up the process or is it according to coding standards?

Evizero: well `zeros()` allocates a new array every time it is called and then fills that array with zeros.
this new approach reuses the same array every time and just resets the values. In pretty much all cases that i encountered this kind of preallocation will be faster

Evizero:
LGTM as far as I can judge

I do think though this needs to be linked to from somewhere. probably https://github.com/JuliaImages/juliaimages.github.io/blob/source/docs/src/demos.md (right @timholy ?)

aquatiko:
@Evizero Yes, I forgot about that. Will add it soon.
I saw on the idea page of **GSoC(Julia)** about adding demos as a project.
I am a bit confused if all the demos needs to be GUI supported or should they just show different things and techniques we can use from JuliaImages?

Evizero:
> I am a bit confused if all the demos needs to be GUI supported or should they just show different things and techniques we can use from JuliaImages?

The later I am pretty sure. Maybe Prof. Holy has something to add to that

Evizero:
can you rebase your commits on the current `source` branch? I made a force push to fix some things

aquatiko:
@Evizero Can this be merged now?

Evizero:
thanks!

PR #47, opened 2019-03-10
Add demonstration: RGB to grayscale

mgautam98:
Hey
I have added a demonstration for converting an image from RGB to grayscale.
I know it is trivial but can you @timholy provide feedback as I want to work on more demos.

I took reference from [scikit-image](http://scikit-image.org/docs/dev/auto_examples/color_exposure/plot_rgb_to_gray.html#sphx-glr-auto-examples-color-exposure-plot-rgb-to-gray-py)

Thank You


timholy:
Hi @mgautam98, sorry to be unresponsive, this arrived when I was swamped with writing the debugger and I've only just noticed it.

I'd love to get a demo on this topic and #48. The framework has changed a lot with DemoCards. I'd also recommend avoiding the usage of ImageView and just focus on generating the images, allowing them to be displayed by DemoCards.

Are you interested in updating this or should I?

PR #48, opened 2019-03-12
Add Demo:Histogram Equalization

mgautam98:
Hey

I have added a demo on histogram equalization. I would like to know your feedback @timholy. 

Thank you


timholy:
If you update this, note that the histogram equalization code has moved to https://github.com/JuliaImages/ImageContrastAdjustment.jl (which is exported by Images, but you don't necessarily need to use Images itself anymore).

PR #56, opened 2019-06-25
Add example images for `imresize`

JeffFessler:
I had left the example images as a "todo" earlier because I wanted to see if the PR was accepted first.


johnnychen94:
Thank you for writing this

I replaced the first `imresize` example using keyword `ratio`, which I think is more concise.

PR #58, opened 2019-08-02
rewrite quickstart

johnnychen94:
Although I'm still not an expert of JuliaImages, I think I'm capable of writing some easy pieces of documentation as a conclusion part of my GSoC project.

I rewrite all the codes in `quickstart.md` to give new users a better intuition on what JuliaImages provides.

What I'm planning to do after this PR:

I plan to move some of the "hard"/"advanced" part to other places, for example, the reasoning behind "The 0-to-1 intensity scale" and "Array elements are pixels". I'm not sure of it, but perhaps most users (especially those from MATLAB) aren't aware of these concepts, instead, these topics are prepared for package developers IMO.

"Some add-on packages enable additional behavior" section might be also quite advanced (at least I haven't got such a chance to dig into it 😂 )

A sample of the built page
image


docs/src/quickstart.md
```diff
@@ -72,29 +111,49 @@ smoothly with the vast majority of functions in JuliaImages.
...
+           RGB(1, 1, 1), # hide
+           nrow=1, npad=2) # hide
 ```
 
-(In ImageView, you can "hover" over pixels and see their coordinates in the statusbar.)
-The meaning of `N0f8` is summarized [below](@ref The-0-to-1-intensity-scale) and described in detail
-[later](@ref fixedpoint).
+As you can see, both `img_rgb` and `img_lab` images are of size
+``2 \times 2`` instead of ``2 \times 2 \times 3`` or ``3 \times 2 \times 2``;
+RGB image has `RGB` pixels, Lab image has `Lab` pixel.
```
zygmuntszpak: a RGB image is an array of `RGB` pixels whereas a Lab image is an array of  `Lab` pixel.

docs/src/quickstart.md
```diff
@@ -72,29 +111,49 @@ smoothly with the vast majority of functions in JuliaImages.
...
+
+!!! note
+    It's recommended to use `Gray` compared to `Number` in JuliaImages since it has a
```
zygmuntszpak: It's recommended to use `Gray` instead of the `Number` type in JuliaImages since it signals that the array of numbers is best interpreted as a grayscale image. For example, it triggers `Atom/Juno` and `Jupyter`  to display the array as an image instead of a matrix of numbers. There's no performance overhead for using `Gray` over `Number`.

docs/src/quickstart.md
```diff
@@ -107,66 +166,130 @@ processing" communities.
...
+```
+
+!!! note
+    You'll see [broadcasting](https://docs.julialang.org/en/v1/manual/arrays/#Broadcasting-1)
+    semantics used in JuliaImages here and there, check the documentation if
+    you're not familiar with it.
 
-```jldoctest
-using Images
+Sometimes, to work with other packages, you'll need to convert a ``m \times n``
```
zygmuntszpak: a ``m \times n`` `RGB` image to an ``m \times n \times 3``  image and vice versa.

docs/src/quickstart.md
```diff
@@ -107,66 +166,130 @@ processing" communities.
...
+`RGB` images to ``m \times n \times 3`` image and vice versa. `channelview` and
+`colorview` are designed for this purpose. For example:
```
zygmuntszpak: The functions  `channelview` and `colorview` are designed for this purpose.

docs/src/quickstart.md
```diff
@@ -107,66 +166,130 @@ processing" communities.
...
-# RGB images
-A = rand(3, 8, 8)            # Float64 data (64 bits per color channel)
-img = colorview(RGB, A)      # encodes as a 2d RGB{Float64} array
-A = rand(N0f8, 3, 8, 8)      # uses only 8 bits per channel
-img = colorview(RGB, A)      # encodes as a 2d RGB{Float64} array
+!!! note
+    The reason we use CHW (i.e., channel-height-width) order instead of HWC
+    is that this provides a memory friendly indexing mechanisim for `Array`.
+    By default, in Julia the first index is also the fastest (i.e., has
+    adjacent storage in memory). For more details, please check [Access arrays in memory order, along columns](https://docs.julialang.org/en/v1/manual/performance-tips/#Access-arrays-in-memory-order,-along-columns-1)
```
zygmuntszpak: For more details, please refer to the performance tip: [Access arrays in memory order, along columns.](https://docs.julialang.org/en/v1/manual/performance-tips/#Access-arrays-in-memory-order,-along-columns-1)

johnnychen94:
> What do you think about the Function categories part? A lot of the functions listed under Algorithms are probably not what one would call anymore. 

Yes, I had the same thought and planned to work on it in the future. I just want to limit myself into quickstart.md in this PR.

timholy:
Great stuff, thanks!

PR #93, opened 2019-12-01
add compat section

johnnychen94:
Only main packages have a minor version compatibility, e.g., `Images`,
`Documenter`

close #66 
close #67 
close #68 
close #69 
close #70 
close #71 
close #72 
close #73 
close #74 
close #75 
close #76 
close #77 
close #78 
close #79 
close #80 
close #81 
close #82 
close #83
close #84
close #85 
close #86 
close #87 
close #88 
close #89 
close #90 
close #91
close #92


timholy:
0.5 is allowed to be breaking compared to 0.4, so this allows version upgrades to break the documentation. Is that what we want?

johnnychen94:
The current compat section says, okay, I'll build the documentation of `Images v0.19" using `Documenter v0.24`, and I don't care about all other versions. The version resolve process should find the latest available and compatible version for all other packages.

timholy:
I see your point. Here your philosophy is to let the other packages take control? That's not unreasonable, in fact it's quite a good idea! Let's merge this.

PR #95, opened 2019-12-03
limit main package versions to patch level

johnnychen94:


johnnychen94:
By doing this, whenever these main packages tag a new release, CompatHelper would send a notification.

timholy:
Once we go to 1.0, patch-level changes should be bugfix only, but for now this is OK.

My comment actually referred to all the other dependencies, which are much more generous in their bounds and thus far less likely to trigger a rebuild.

PR #97, opened 2019-12-08
remove outdated ImageView warning

johnnychen94:
@timholy With ImageView `0.10` released, I assume this is outdated now? Or, are OSX and Windows only works on Julia 1.3?


timholy:
Yes, it can be removed! Though really only Julia 1.3 is "safe," earlier versions will likely encounter problems on some platforms.

PR #99, opened 2019-12-13
replace travis with github action

johnnychen94:
Before we figure out what happens to #98, another option is to deploy the docs using Github Action


johnnychen94:
Hmmm, seems that julia 1.0 on github action is also affected 😕

timholy:
Likely also noted in https://travis-ci.org/JuliaImages/Images.jl/jobs/624338186?utm_medium=notification&utm_source=github_status

timholy:
This is not a JuliaImages problem, it's a Julia problem. I've been able to replicate this locally with julia-1.0.5 (but not julia-1.0.0), and am running a `git bisect` now.

timholy:
https://github.com/JuliaLang/julia/issues/34121. Probably a WIP.

PR #102, opened 2019-12-28
Update ImageMetadata documentation

Tokazama:
This updates documentation to reflect the changes that will occur from merging in https://github.com/JuliaImages/ImageMetadata.jl/pull/58.

I should probably mention that I was able to successfully build the portion of the documentation related to this PR locally (with the appropriate changes to ImageMetadata.jl in place locally).


johnnychen94:
Great, I think the best time to merge this PR is when Images.jl set its compatibility to the ImageMetadata v0.8.0 (currently it's v0.7.2), some delay there is expected since some efforts are still needed to clean up the dependencies from ColorTypes.jl all the way up to Images.jl

PR #106, opened 2020-01-12
Add a demonstration for detecting contours

archit120:
As discussed on Slack. The script detects contours in binary images and returns a list of list of points (similar to openCV). The algorithm used is "Topological Structural Analysis of Digitized Binary Images by Border Following" by Suzuki and Abe (Same as OpenCV). 

My implementation however doesn't preserve hierarchy. This can be added easily.

### DIFF
--- a/docs/examples/contours/contour_detection.jl
+++ b/docs/examples/contours/contour_detection.jl
@@ -1,0 +1,168 @@
+# ---
+# cover: assets/contour_detection.png
+# title: Contour Detection and Drawing
+# ---
+
+# This demonstration shows how to detect contours on binary images
+# the algorithm implemented is exactly same as OpenCV
+
+using Images, TestImages, FileIO
+dir_delta = [[-1, 0] , [-1, 1], [0, 1], [1, 1], [1, 0], [1, -1], [0, -1], [-1,-1]]
+
+function clockwise(dir)
+    return (dir)%8 + 1
+end
+
+function counterclockwise(dir)
+    return (dir+6)%8 + 1
+end
+
+
+function move(pixel, image, dir)
+    newp = pixel + dir_delta[dir]
+    if newp[1] > 0 && newp[1] <= size(image, 1) && newp[2] > 0 && newp[2] <= size(image, 2)
+        if image[newp...]!=0
+            return newp
+        end
+    end
+    return [0,0]
+end
+
+
+function from_to(from, to)
+    delta = to-from
+    return findall(x->x == delta, dir_delta)[1]
+end
+
+
+
+function detect_move(image, p0, p2, nbd, border)
+    dir = from_to(p0, p2)
+    moved = clockwise(dir)
+    p1 = [0, 0]
+    while moved != dir
+        newp = move(p0, image, moved)
+        if newp[1]!=0
+            p1 = newp
+            break
+        end
+        moved = clockwise(moved)
+    end
+
+    if p1 == [0,0]
+        return
+    end
+
+    p2 = p1
+    p3 = p0
+
+    done = [false, false, false, false, false, false, false, false]
+    while true
+        dir = from_to(p3, p2)
+        moved = counterclockwise(dir)
+        p4 = [0,0]
+        done = [false, false, false, false, false, false, false, false]
+        while true
+            p4 = move(p3, image, moved)
+            if p4[1] != 0
+                break
+            end
+            done[moved] = true
+            moved = counterclockwise(moved)
+        end
+        push!(border, p3)
+        if image[p3...] != 0 && (p3[1] == size(image, 1) || done[5])
+            image[p3...] = -nbd
+        elseif image[p3...] == 1
+            image[p3...] = nbd
+        end
+
+        if (p4 == p0 && p3 == p1)
+            break
+        end
+        p2 = p3
+        p3 = p4
+    end
+end
+
+function find_contours(image)
+    nbd = 1
+    lnbd = 1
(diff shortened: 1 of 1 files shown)


johnnychen94:
The results look great to me:

image

Can you also add some comments here so that others can understand it? Adding the reference to the page would also be very helpful.

archit120:
Comments in between functions seem to break the documentation generator. This is partly why I removed all the comments. Could you look into it? I'll add back the comments once that works.

johnnychen94:
using [Literate's syntax](https://fredrikekre.github.io/Literate.jl/stable/fileformat/), comments should be leading two `#`s (no spaces between them)

```julia
# # This is a markdown level-1 title

# this is rendered as markdown contents

function foo()
    ## this is rendered as julia source code comment
    println("hello world")
end
```

The advanced filter syntax such as leading `#md` seems broken at present. I need to find some to fix `DemoCards.jl`.

archit120:
I've added some comments. Do I need to add more or will this suffice?

johnnychen94:
It looks very nice when I build it locally. Regardless of the naming issues (e,g., `dir`, `from`, `to`), this can be a pretty good demo already. I'm not familiar with this topic so I didn't check the implementation details. 

Before merging this demo, I'll wait for @zygmuntszpak for several days to see if he's available to check this.

P.S. you could add your name to this demo as well if you're proud of it :P

zygmuntszpak:
Hi Archit,

I was just about to look into implementing the  _Topological Structural Analysis of Digitized Binary Images by Border Following_ paper, so your time timing is impeccable. I intend to carve out some time in the next couple of days to read the paper more closely and think about your implementation, and how we might make it more performant etc.

archit120:
Hi @zygmuntszpak,

I'd really appreciate any feedback about this implementation. Thanks for taking out the time to look at this!

docs/examples/contours/contour_detection.jl
```diff
@@ -0,0 +1,184 @@
...
+                    from[1] += 1
+                end
+
+                p0 = [i,j]
+                detect_move(image, p0, from, nbd, border) ## 3
+                if size(border)==0
+                    append!(border, p0)
+                    image[p0...] = -nbd
+                end
+                unique_map[sort(border)]=1
```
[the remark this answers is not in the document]
archit120: I was having a problem with duplicate contours being detected. It might, however, have been a consequence of some flawed implementation and I didn't remove this. I will try again without this.

johnnychen94:
I prefer to merge this PR first, then update the demo when the functionality is implemented somewhere, which might take a long period. By doing this we can progressively have something informative to the community, @zygmuntszpak how do you think?

zygmuntszpak:
Since others might look at the demo as an example of how to code in Julia, I would prefer if there wasn't so much needless memory allocation just for indexing into parts of the image. On the other hand, I agree that this particular demo will probably be replaced in the near future once we have a solid and tested implementation of this particular algorithm in a package.

 @archit120 Do you envisage the type of changes I suggested to take you a long time to do? 

I agree that we don't need an ideal implementation for this demo, because having something is better than having nothing. I don't intend to act as a roadblock so merge if you want, but my preference is for a somewhat "cleaner" implementation.

archit120:
@zygmuntszpak I have a bit of a time crunch this week because of assignments. I'll still try to get it done in the next couple of days.

archit120:
Hey,

Thanks for the comments. So there was a tiny implementation mistake. I fixed that and also made the recommended changes. This is on the original implementation:
`4.912141 seconds (80.28 M allocations: 2.810 GiB, 19.77% gc time)`
By implementing changes mentioned in comments:
`1.311359 seconds (18.42 M allocations: 685.099 MiB, 12.82% gc time)`
Finally, after fixing the implementation bug:
'0.189207 seconds (789.08 k allocations: 39.890 MiB)'

johnnychen94:
Looks nice. I'll call it a great demo here and merge this in two days if there's nothing else to complement

docs/examples/contours/contour_detection.jl
```diff
@@ -0,0 +1,192 @@
...
+# This demonstration shows how to detect contours on binary images
+# The algorithm used is ["Topological Structural Analysis of Digitized Binary Images
+# by Border Following"](https://www.sciencedirect.com/science/article/pii/0734189X85900167) by Suzuki and Abe (Same as OpenCV).#
+
+# Points are represented using CartesianIndex. Contours are represented as a vector of Points.
+# Direction is a single number between 1 to 8. Steps inside functions are marked as they are in the original paper
+
+using Images, TestImages, FileIO
+
+##              N          NE      E       SE      S       SW        W      NW
```
zygmuntszpak: Probably don't need this comment here anymore.

zygmuntszpak:
Fantastic, well done on obtaining such a great performance improvement!

docs/examples/contours/contour_detection.jl
johnnychen94: I think they still make sense here; explaining the inputs of `clockwise` and `counterclockwise`, so I'll keep them unchanged.

johnnychen94:
Hmmm, the next step is to find a place for this algorithm. @zygmuntszpak I've forgotten the detailed plan you told me in slack and the records are missing. Can you repost them here so that @archit120 knows what he needs to do next?

archit120:
@johnnychen94 @zygmuntszpak what should I do next?

zygmuntszpak:
I've posted to the thread on the Slack channel.

PR #110, opened 2020-02-08
Add a package dashboard

timholy:
This uses the new https://github.com/baggepinnen/PkgDashboards.jl. You can see it here: https://github.com/JuliaImages/juliaimages.github.io/blob/teh/dashboard/DASHBOARD.md


johnnychen94:
FYI, 5accda3 updates some badges accordingly for Github Actions and coverall

PR #114, opened 2020-03-05
use DemoCard v0.2

johnnychen94:
A lot of tweaks and bug fixes on DemoCards since v0.1, also the codes related to themes are rewritten with a breaking API, which will be stable by now IMO.


timholy:
Awesome. I only just noticed that several demos are queued here (again having arrived while I was focusing on the debugger). Should we look into getting them merged, perhaps doing the porting for them? I'd be happy to do much of it as a way of familiarizing myself with DemoCards.

johnnychen94:
Codes in #47 and #48 looks trivial to me, but rewrite them with a heavy explanation on the JuliaImages sounds like a good tutorial for new users. We should be proud of the color systems and make good advertisement on it :P

timholy:
I decided that out of fairness we should give the authors a chance to update them; if they don't respond soon I'll polish them up myself.

PR #121, opened 2020-04-10
set travis timeout 20min

johnnychen94:
The previous documentation building CI fails because it exceeds the default timeout 10min, let's see how things work out if we set it to 20min.


timholy:
I'm seeing an error when I try to fetch the log. You too?

johnnychen94:
Yes, it's a little weird here.

johnnychen94:
It seems like building docs using Julia 1.3 works.

close in favor of #123

PR #123, opened 2020-04-15
Various updates

timholy:
This fixes quite a lot of the errors/warnings that get emitted when we build the docs. There are still many warnings about "potentially missing docstrings," but a lot of these are for internal functions (https://github.com/JuliaDocs/Documenter.jl/issues/600) or things that ColorTypes/Colors exports and which probably don't need to be documented here.

Some failures I have already fixed locally and will be fixed by the next releases of ColorTypes, PaddedViews, ImageMetadata, ImageSegmentation, and ImageContrastAdjustment.

There are also a few remaining oddities. @zygmuntszpak, for some reason it's not finding the docs for ImageContrastAdjustment. @johnnychen94, for some reason it's not finding the docs for `ssd`, `mse`, `sad`, and `mae`. I poked at these briefly but didn't quickly find the reason.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: julia
 os:
   - linux
 julia:
-  - 1.0
+  - 1.3
 env:
   - DOCUMENTER_DEBUG=true
 notifications:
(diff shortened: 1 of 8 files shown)


timholy:
I should say that locally the main remaining errors are:

- CoordinateTransformations now uses a different precision in printing
- ImageSegmentation seems to have changed internals and we should check and rebuild the docs here if everything seems OK.

Aside from the missing docstrings I mentioned above, that's it! Rest are fixed locally (but the Travis run won't show that just yet).

johnnychen94:
I kept finding it necessary to fix these but every time when I tried, I found myself not getting that level of a comprehensive picture of the available functions. Thanks for doing this! 

> for some reason it's not finding the docs for ssd, mse, sad, and mae

Will check that later in a separate PR.

PR #137, opened 2020-06-29
Clarified use

pauljurczak:
`using Images` is necessary for examples on the following pages to work, but it's not shown explicitly here or on the next page. See also https://discourse.julialang.org/t/efficient-way-to-find-which-package-defines-an-unknown-type/42222

### DIFF
--- a/docs/src/install.md
+++ b/docs/src/install.md
@@ -49,7 +49,7 @@ Adding these gives you a basic image IO setup:
 and to load an image, you can use
 
 ```@example
-using FileIO
+using Images, FileIO
 using ImageShow # hide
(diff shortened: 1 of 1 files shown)


johnnychen94:
Thanks, this was written such way that `using Images` is not needed to load an image. But since you feel this confusing, I'm happy to get this in.

pauljurczak:
Thanks. I realize that it wasn't necessary for the first page, but it may be helpful for readers going page by page, copying and executing code snippets from the manual. Just an extra safety margin for newbies.

timholy:
Thanks Paul!

PR #138, opened 2020-06-30
add demo: inpainting

mgautam98:
Added a demo of Image inpainting
@johnnychen94

### DIFF
--- a/docs/examples/filters_and_restorations/inpainting.jl
+++ b/docs/examples/filters_and_restorations/inpainting.jl
@@ -1,0 +1,42 @@
+# ---
+# title: Image Inpainting
+# cover: assets/inpainting.png
+# ---
+
+# This demo illustrates the use of [image inpainting](https://github.com/JuliaImages/ImageInpainting.jl) algorithm in JuliaImages.
+
+# Inpainting is the process of constructing lost or deteriorated parts of an
+# image or video. It is also used for removing undesired elements from a scene.
+#  
+# The basic idea of inpainting is: Replace those bad marks (or undesired elements) with its 
+# neighbouring pixels so that it looks like the neighbourhood.
+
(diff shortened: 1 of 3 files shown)


docs/examples/filters_and_restorations/inpainting.jl
```diff
@@ -0,0 +1,42 @@
...
+# image or video. It is also used for removing undesired elements from a scene.
+#  
+# The basic idea of inpainting is: Replace those bad marks (or undesired elements) with its 
+# neighbouring pixels so that it looks like the neighbourhood.
+
+# In this demo we will be removing an undesired element form an image.
+
+using ImageInpainting, TestImages, ImageCore
+
+img  = Float64.(Gray.(testimage("lighthouse")));
```
johnnychen94, line 18: Is it because the `inpaint` only accepts numerical array as inputs? In that case, we should fix and make it also works for `AbstractArray{<:AbstractGray}`.

mgautam98: Okay I would like to work on that

johnnychen94: It looks like the algorithm isn't optimized well, the memory allocation is conspicuous to me.

```julia
julia> @time out = inpaint(img, mask, Criminisi(30,30));
 14.502141 seconds (2.16 M allocations: 13.540 GiB, 18.97% gc time)
```

I'm not sure what's on your TODO list, but I guess this is also worth checking.

ashwani-rathee: Update on this: 
```julia
@time out = inpaint(img, mask, Criminisi(30,30));
 17.644576 seconds (4.94 M allocations: 13.659 GiB, 3.46% gc time)
```
😄 It's pretty slow,let's check

PR #142, opened 2020-06-30
Made an example more illustrative

pauljurczak:
There is an issue of different canonical image orientations, e.g. https://discourse.julialang.org/t/different-image-orientation-in-images-vs-makie/42288. Making a matrix in this example non-square will alert the reader that Julia orients images (matrices) differently than some other programming environments.


timholy:
You are a dream user, @pauljurczak! Thanks for these improvements!

PR #144, opened 2020-07-01
demos: add imcorner demo

jiviteshjain:
This adds a demo for corner detection. Not sure if it should go in the "Contour" category. @johnnychen94

### DIFF
--- a/docs/examples/contours/detecting_corners.jl
+++ b/docs/examples/contours/detecting_corners.jl
@@ -1,0 +1,68 @@
+# ---
+# cover: assets/corner.png
+# title: Detecting Corners
+# ---
+
+# Corner Detection, which is a subset of Interest Point Detection, tries to detect
+# points in the image which have a well-defined position and can be robustly detected
+# in multiple images of the same scene. Very often, these points lie along the corners
+# or edges of objects in the image - hence the name.
+
+# Corner detection is useful in several computer vision tasks - such as Image Registration,
+# Motion Detection and Panaroma Stitching. This is because if the locations of the same
+# points are known in two different images, it gives a reference to align those images. Corners,
+# with their well-defined positions serve as good candidates for such points.
+
+using Images, TestImages
+img = Gray.(testimage("house"))
+
+# We use this image of a house, with numerous edges and corners.
+
+# The `imcorner` function can be used to detect corners in the image - and it returns an
+# array of booleans, where a `true` value denotes that the corresponding pixel may be a
+# corner. We use this to mark those pixels in red on a copy of the image.
+
+corners = imcorner(img)
+img_copy = RGB.(img)
+img_copy[corners] .= RGB(1.0, 0.0, 0.0)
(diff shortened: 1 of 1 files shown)


docs/examples/contours/detecting_corners.jl
```diff
@@ -0,0 +1,68 @@
...
+# in multiple images of the same scene. Very often, these points lie along the corners
+# or edges of objects in the image - hence the name.
+
+# Corner detection is useful in several computer vision tasks - such as Image Registration,
+# Motion Detection and Panaroma Stitching. This is because if the locations of the same
+# points are known in two different images, it gives a reference to align those images. Corners,
+# with their well-defined positions serve as good candidates for such points.
+
+using Images, TestImages
+img = Gray.(testimage("house"))
```
johnnychen94, line 17: It's totally fine here but I'd like to add a side note to this.

It works as expected here because it can infer the storage type to be `N0f8`, but for array allocation functions (e.g., `zeros`, `Array{Gray, 2}(undef, 4, 4)`) there will be some significant performance difference:

```julia
julia> @btime zeros(Gray, 100, 100);
  39.331 μs (10002 allocations: 234.45 KiB)

julia> @btime zeros(Gray{N0f8}, 100, 100);
  2.565 μs (1 allocation: 9.94 KiB)

julia> @btime Array{Gray, 2}(undef, 100, 100);
  1.841 μs (2 allocations: 78.20 KiB)

julia> @btime Array{Gray{N0f8}, 2}(undef, 100, 100);
  127.157 ns (1 allocation: 9.94 KiB)
```

I'm not the expert of this , but I think this is because its `eltype`  `Gray{Any}` so array allocation can't infer the best layout, and would make everything later much slower. @kimikage Am I understanding this correctly here?

kimikage: Your understanding is (mostly) correct, but I don't think it's a good place to write the explanation here.

timholy: I would say that here we can rely on inference to do the right thing---we've worked hard in ColorTypes to make it work. Let's save that explanation for some place where it is necessary.

docs/examples/contours/detecting_corners.jl
```diff
@@ -0,0 +1,68 @@
...
+end
+mosaicview(img_copies; nrow=1)
+
+# These algorithms use the gradient of the image to identify corners, because intensities
+# change abruptly at corner points, giving rise to large gradients. However, this makes them
+# computationally expensive.
+
+# The *FAST (Features from Accelarated Segment Test) Corner Detector* is a feature
+# detection algorithm which is designed to be computationally cheaper, and hence much faster.
+# It classifies a pixel $P$ as a corner if atleast $n$ contiguous points out of the 16
```
timholy, line 61:
```suggestion
# It classifies a pixel $P$ as a corner if at least $n$ contiguous points out of the 16
```

Issue #139, opened 2020-06-30
document the difference between float32 and Float32, etc.


johnnychen94:
We need to advertise the usage of lower-cased form as it's simpler and more generic:

```julia
float32.(img) # promoting the storage type while preserving its colorant

Gray{Float32}.(img) # this could accidentally convert an RGB image to Gray image
```

Ref: https://github.com/JuliaImages/ImageQualityIndexes.jl/pull/20

PR #146, opened 2020-07-01
document the difference between float32 and Float32, etc.

mgautam98:
Solves #139 
Added documentation to advertise the usage of lower-cased forms of Float32, Float64 etc


docs/src/tutorials/conversions_views.md
```diff
@@ -507,3 +507,33 @@ Base.ReshapedArray{RGB{Normed{UInt8,8}},3,Base.ReinterpretArray{RGB{Normed{UInt8
 While there is little or no performance cost to making use of
 JuliaImage's convenient views, sometimes the types can get
 complicated!
+
+## Preserving colorant type while conversion
+
+While using Float32, Float64 to convert the raw storage type of an image
```
johnnychen94, line 513: Let's quote these types with `

docs/src/tutorials/conversions_views.md
```diff
@@ -507,3 +507,33 @@ Base.ReshapedArray{RGB{Normed{UInt8,8}},3,Base.ReinterpretArray{RGB{Normed{UInt8
 While there is little or no performance cost to making use of
 JuliaImage's convenient views, sometimes the types can get
 complicated!
+
+## Preserving colorant type while conversion
```
timholy, line 511: Let's instead put this info in `arrays_colors.md` right before the `A consistent scale for floating-point and "integer" colors: fixed-point numbers` section.

docs/src/tutorials/conversions_views.md
```diff
@@ -507,3 +507,33 @@ Base.ReshapedArray{RGB{Normed{UInt8,8}},3,Base.ReinterpretArray{RGB{Normed{UInt8
...
+
+While using Float32, Float64 to convert the raw storage type of an image 
+the color space information is lost. Replacaing Float32, Float64 with their lower-case
+forms preserves the colorant type.
+
+Take an example of converting a Gray image to Float64. Simply broadcating Float64
+will promote the storage to Float64 but the color space information is lost.
+However, when we use float64. The colorant type inforamtaion is preserved.
```
johnnychen94, line 519: This paragraph seems not as illustrative as the following codes so I suggest removing it.

johnnychen94, line 519: Another major difference between `Float64` and `float64` is that `Float64.(rgb_img)` will error while the `float64.(rgb_img)` one works as expected.

docs/src/tutorials/conversions_views.md
```diff
@@ -507,3 +507,33 @@ Base.ReshapedArray{RGB{Normed{UInt8,8}},3,Base.ReinterpretArray{RGB{Normed{UInt8
...
+
+julia> Float64.(img) |> eltype
+Float64
+
+julia> float64.(img) |> eltype
+Gray{Float64}
+```
+|  Storage Type 	|  Alternative 	|
+|---	|---	|
+| Float32  	|  float32 	|
```
johnnychen94, line 533: Let's quote these, too. And it would be better to quote the lower-cased ones with ref-links:

```
[`float32`](@ref)
```

docs/src/tutorials/conversions_views.md
```diff
@@ -507,3 +507,33 @@ Base.ReshapedArray{RGB{Normed{UInt8,8}},3,Base.ReinterpretArray{RGB{Normed{UInt8
 While there is little or no performance cost to making use of
 JuliaImage's convenient views, sometimes the types can get
 complicated!
+
+## Preserving colorant type while conversion
+
+While using Float32, Float64 to convert the raw storage type of an image 
+the color space information is lost. Replacaing Float32, Float64 with their lower-case
+forms preserves the colorant type.
```
timholy, line 515: I think it would be better to present this as an example of how you control color space and element type generally. There are two orthogonal concepts:

- what colorspace do you want: `RGB`, `HSV`, `Gray`, or pure numeric?
- what precision of storage type do you want: `Float32`, `N0f8`, etc.?

The lowercase `float32`, `n0f8` are designed to control the second *without modifying the first*. In contrast, `Gray.(img)` and `HSV.(img)` (*without* the eltype specification) control the first without (as far as possible) modfiying the second. And finally, `Gray{N0f8}.(img)` and `HSV{Float32}.(img)` (*with* the eltype specification) control both simultaneously.

Limiting your explanation to grayscale also misses out on the opportunity to present the broader perspective on color spaces in general, and might make the distinction between `float32` and `Float32` seem like hair-splitting. The importance of the distinction becomes much clearer in the context of color images. I'd also recommend demoing the difference between `float32` and `Float32` applied to an RGB image: while for grayscale you have
```julia
julia> img = testimage("cameraman");

julia> summary(img)
"512×512 Array{Gray{N0f8},2} with eltype Gray{Normed{UInt8,8}}"

julia> summary(float32.(img))
"512×512 Array{Gray{Float32},2} with eltype Gray{Float32}"

julia> summary(Float32.(img))
"512×512 Array{Float32,2}"
```
for a color image you have
```julia

julia> summary(img)
"512×512 Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}"

julia> summary(float32.(img))
"512×512 Array{RGB{Float32},2} with eltype RGB{Float32}"

julia> summary(Float32.(img))
ERROR: MethodError: no method matching Float32(::RGB{Normed{UInt8,8}})
Closest candidates are:
  Float32(::Float64) at float.jl:251
  Float32(::Int16) at float.jl:60
  Float32(::UInt16) at float.jl:66
  ...
Stacktrace:
[...]
```
That's because there's no unique way to convert a `RGB` to `Float32`. However, we may be being a bit pedantic here, because there is an [agreed-upon conversion to gray](https://stackoverflow.com/questions/687261/converting-rgb-to-grayscale-intensity), and we use it:
```julia
julia> summary(Gray{Float32}.(img))
"512×512 Array{Gray{Float32},2} with eltype Gray{Float32}"

julia> summary(Float32.(Gray{Float32}.(img)))
"512×512 Array{Float32,2}"
```

I think the reason we don't use it is because some people might expect `Float32.(img)` to do the thing that `float32.(img)` does.

docs/src/tutorials/conversions_views.md
mgautam98: why does `Float64.(rgb_img)` throws error?

timholy:
```julia
julia> using Colors

julia> c = Gray(0.5)
Gray{Float64}(0.5)

julia> Float64(c)
0.5

julia> c = RGB(0.2, 0.4, 0.8)
RGB{Float64}(0.2,0.4,0.8)

julia> Float64(c)
ERROR: MethodError: no method matching Float64(::RGB{Float64})
Closest candidates are:
  Float64(::BigFloat) at mpfr.jl:332
  Float64(::BigFloat, ::Base.MPFR.MPFRRoundingMode) at mpfr.jl:332
  Float64(::BigFloat, ::RoundingMode) at mpfr.jl:334
  ...
Stacktrace:
 [1] top-level scope at REPL[5]:1
```
Tell me what `Float64` value should `Float64(RGB(0.2, 0.4, 0.8))` return? There are 3 color channels.

My comment below points out it's a little more complicated because there is an agreed-upon standard for converting RGB to a pure grayscale intensity. So
```julia
julia> Float64(Gray(c))
0.38580000400543213
```
The question is whether `Float64(c)` should do that on its own. The concern is that too many people will want to to mean element-wise conversion, which you instead should do this way:
```julia
julia> RGB{Float32}(c)
RGB{Float32}(0.2f0,0.4f0,0.8f0)
```
or in shorthand
```julia
julia> using ImageCore

julia> float32(c)
RGB{Float32}(0.2f0,0.4f0,0.8f0)
```

JuliaImages tries not to read your mind. We ask programmers to be clear about what they want, otherwise subtle bugs are more likely and one always loses a lot of time to tracking them down.

mgautam98: Thanks for explaining.

timholy: Glad it helped. And to be sure that the more general principle is also clear: if `T` is a type, `T(x)` should return an object of type `T`. Since `Float64` is a type, it needs to return a single `Float64` object. But `float64` is not a type (it's a function), so it's under no such constraints. So we made it operate on the "storage type" without modifying the "color space." That's what I meant below about "orthogonality," essentially keeping the two concepts separate from one another.

docs/src/tutorials/conversions_views.md
johnnychen94: > I think the reason we don't use it is because some people might expect Float32.(img) to do the thing that float32.(img) does.

Yes, it would be awkward and newbies-unfriendly that `Float32.(rgb_img)` returns a different result from `float32.(rgb_img)` without warning. This can then be the root of many hard-to-discover bugs.

docs/src/tutorials/arrays_colors.md
```diff
@@ -222,6 +222,57 @@ These values are close to the channels of `c`, but have been rounded
...
+Closest candidates are:
+  Float32(::Int8) at float.jl:60
+  Float32(::Int16) at float.jl:60
+  Float32(::Int32) at float.jl:60
+  ...
+  Stacktrace:
+  [...]
+```
+**Note**: `Float64.(rgb_img)` throws error because it does not implicitly   
+convert RGB to single channel but `float64.(rgb_img)` works as expected.
```
timholy, line 263: "Works as expected" depends entirely on what your expectations are. Some people will expect it to return a single floating-point number.

How about copy/pasting the examples with a single colorant object that I gave in my explanation to you? This mixes together broadcasting with colorspace conversion, and we don't have to mix the concepts.

johnnychen94:
I'd like to also add the following table into the document as a cheat sheet:

The following table gives you how you can convert from a number/colorant to another number/colorant

| from \ to        | `N0f8`           | `Float32`          | `Gray{N0f8}`    | `Gray{Float32}`       | `RGB{N0f8}`       | `RGB{Float32}`    |
| -----------      | ------------     | --------           | -----------     | --------              | ---------         | --------          |
| `N0f8`           |    -             | `float32(c)`       | `Gray(c)`       | `Gray{Float32}(c)`    | `RGB(c)`          | `RGB(c)`          |
| `Float32`        |  `N0f8(c)`       | -                  | `Gray{N0f8}(c)` | `Gray(c)`             | `RGB{N0f8}(c)`    | `RGB(c)`          |
| `Gray{N0f8}`     |  `N0f8(c)`       | `Float32(c)`       |  -              | `float32(c)`          | `RGB(c)`          | `RGB{Float32}(c)` |
| `Gray{Float32}`  |  `N0f8(c)`       | `Float32(c)`       | `n0f8(c)`       | -                     | `RGB{N0f8}(c)`    | `RGB(c)`          |
| `RGB{N0f8}`      |  `N0f8(Gray(c))` | `Float32(Gray(c))` | `Gray(c)`       | `Gray{Float32}(c)`    | -                 | `float32(c)`      |
| `RGB{Float32}`   |  `N0f8(Gray(c))` | `Float32(Gray(c))` | `Gray{N0f8}(c)` | `Gray(c)`             | `RGB{N0f8}(c)`    | -                 |

* `float32(RGB(c))` works equivelently to `RGB{Float32}(c)`, and their performances are the same, too.
* the lowercase function also works if `c` is a type, for example, `float32(Gray{N0f8})` gives you `Gray{Float32}`
* To handle image data, you just need to broadcast it, for example, `float32.(img)`

Or maybe a simplified version by removing the explicit construction:

| from \ to        | `N0f8`           | `Float32`          | `Gray{N0f8}`    | `Gray{Float32}`       | `RGB{N0f8}`       | `RGB{Float32}`    |
| -----------      | ------------     | --------           | -----------     | --------              | ---------         | --------          |
| `N0f8`           |    -             | `float32(c)`       | `Gray(c)`       | -                     | `RGB(c)`          | `RGB(c)`          |
| `Float32`        |  `n0f8(c)`       | -                  | -               | `Gray(c)`             | -                 | `RGB(c)`          |
| `Gray{N0f8}`     |  -               | -                  | -               | `float32(c)`          | `RGB(c)`          | -                 |
| `Gray{Float32}`  |  -               | -                  | `n0f8(c)`       | -                     | -                 | `RGB(c)`          |
| `RGB{N0f8}`      |  `N0f8(Gray(c))` | `Float32(Gray(c))` | `Gray(c)`       | -                     | -                 | `float32(c)`      |
| `RGB{Float32}`   |  `N0f8(Gray(c))` | `Float32(Gray(c))` | -               | `Gray(c)`             | -                 | -                 |


@timholy will the usage of `RGB(1.0)` being deprecated? I believe it will still be available?

mgautam98:
@johnnychen94 this cheat-sheet is really helpful.

PR #133, opened 2020-06-22
Added demos: Histogram Equalization and RGB to Grayscale conversion

mgautam98:
Added the following demos
1. histogram (equalization, adaptive equalization, gamma correction)
2. RGB to grayscale conversion

Since build was failing I couldn't test it.
#47 #48  Are not needed anymore

PR #147, opened 2020-07-01
fix the cover of rgb2gray demo

johnnychen94:
Since #138 is going to be pended for a while, I give a quick fix here to #133

I instead use gif instead of png so that readers get a better view of the conversion from rgb to gray.

cc: @mgautam98


mgautam98:
Great. Thanks.

PR #149, opened 2020-07-01
update doctest reference for Rotation and Clustering

johnnychen94:
Together with https://github.com/JuliaImages/ImageSegmentation.jl/pull/52 and https://github.com/JuliaImages/ImageMorphology.jl/pull/30 this fixes all doctest warnings when building the juliaimages documentation.

The changes in rotations seem okay. But I'm not very sure if the changes in ImageSegmentation are expected.


timholy:
Segmentation is often at a knife's edge, small changes can lead to differences that affect a lot of individual pixels but don't change the big picture all that much.

PR #150, opened 2020-07-01
add warp example in ImageTransformations

mgautam98:
Added warp example under ImageTransformations. I felt that ImageTransformations was missing example on warp.


johnnychen94:
Weird that CI doesn't get triggered, but it looks good so I'll just merge this.

PR #153, opened 2020-08-26
Discourage `convert(Array{C}, img)` in favor of broadcasting

timholy:
Consistent with https://github.com/JuliaImages/ImageCore.jl/pull/131

### DIFF
--- a/docs/examples/contours/contour_detection.jl
+++ b/docs/examples/contours/contour_detection.jl
@@ -97,7 +97,7 @@ end
 function find_contours(image)
     nbd = 1
     lnbd = 1
-    image = convert(Array{Float64}, image)
+    image = Float64.(image)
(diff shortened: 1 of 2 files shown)


timholy:
Several of these were not problematic in their original form, but my thought is that we should discourage anything that might encourage users to use `convert(Array{C}, img)`.

Do you think we need to have a warning that `convert(Array{RGB}, img)` will result in an abstract eltype but that `RGB.(img)` will be concrete?

johnnychen94:
> Do you think we need to have a docs warning that convert(Array{RGB}, img) will result in an abstract eltype but that RGB.(img) will be concrete?

I can add it in #146. @mgautam98 do you want to continue that PR recently?

mgautam98:
> I can add it in #146. @mgautam98 do you want to continue that PR recently?

Yes, I want to continue. I will add this warning.

PR #125, opened 2020-04-18
reorganize and update documentation

johnnychen94:
The main change of this PR is to reorganize the documentation by splitting the manual into two parts: tutorials and package documentations. Files and assets are moved around and renamed because of this.

Several other changes:

* [index.md] replace documentation TOC with packages list
* [install.md]
  * advertise `mosaicview` as recommended method to display multiple images
  * remove the outdated troubleshooting page as image IO and display are no longer an issue.
* [api_comparison.md] [function_reference.md]
  * update functions provided by ImageDistances and ImageQualityIndexes
  * rewrite "Image loading and saving" section
  * add TOC in function references
* [demos] fixes warnings when image value range exceeds [0, 1]
* remove the duplicated documentation to [ImageFiltering.jl](https://juliaimages.org/ImageFiltering.jl/stable/)

Preview: https://juliaimages.org/previews/PR125

PR #155, opened 2020-09-14
upgrade to Documenter 0.25 and DemoCards 0.3

johnnychen94:
DemoCards v0.3 has brought many new features that I'm eager to upgrade it 😆 It is not a breaking version, the previous build workflow still works, but I'd like to push it forward to the better version.

Author and date meta is added to each demos, which would insert two new badges:

image

There are two broken jldoctest, I skipped the fragile one and fix the other one in https://github.com/JuliaMath/FixedPointNumbers.jl/pull/229

closes #140 
closes #156


johnnychen94:
The next step is to collect the scattered documentation around JuliaImages: https://github.com/JuliaImages/juliaimages.github.io/pull/125#issuecomment-617157921

The path I have in mind is:

* first wrap the current "packages" part docs with the themeless version of `DemoPage` as a proxy, and then
* add `RemoteDemoSection` and `RemoteDemoPage` in DemoCards to support docs in remote repo (e.g., `ImageFiltering`)

Whether this goes well is not checked, though.

PR #158, opened 2020-09-20
sujoy edge demo

amarbit:
Hi 
I added the sujoy edge demo. I checked locally It was working as expected. Kindly accept it.


docs/examples/sujoy_edge/sujoy_edge_demo.jl
```diff
@@ -0,0 +1,80 @@
+# ---
+# cover: assets/sedge.png
+# title: Sujoy Edge Demo
+# ---
+
+using Images,Statistics,ImageView,TestImages
```
johnnychen94, line 6: No need for `ImageView` here.

```suggestion
using Images, Statistics, TestImages
```

docs/examples/sujoy_edge/sujoy_edge_demo.jl
```diff
@@ -0,0 +1,80 @@
+# ---
+# cover: assets/sedge.png
+# title: Sujoy Edge Demo
```
johnnychen94, line 3: Would you mind to give a more meaningful title here? For example, it could be "Image edge detection using Sujoy kernel".

Also, it could be nice to provide a `author` and `date` meta here. For example:

```text
# author: Sujoy Kumar Goswami; Partha Ghosh
# date: 2020-09-21
```

docs/examples/sujoy_edge/sujoy_edge_demo.jl
```diff
@@ -0,0 +1,80 @@
...
+## the Sujoy's filter(s), but does not threshold (default is False).
+##
+## Returns
+## -------
+## edges : gray image (if just_filter is true),
+## else, bool image.
+
+
+
+function sujoy(img, kernel_nhood=0, just_filter=false)
```
johnnychen94, line 34: In this case, let's just use the standard Julia docstring format. https://docs.julialang.org/en/v1/manual/documentation/

docs/examples/sujoy_edge/sujoy_edge_demo.jl
```diff
@@ -0,0 +1,80 @@
...
+    img_channel = channelview(img)
+
+    if size(img_channel)[1] == 3
+        img = Gray.(img)
+        img_channel = channelview(img)
+    end
```
johnnychen94, line 40: The first six lines can be simply replaced by this:

```suggestion
img_channel = Gray.(img)
```

This is better than manually check the first channel.

docs/examples/sujoy_edge/sujoy_edge_demo.jl
```diff
@@ -0,0 +1,80 @@
...
+    max_val = maximum(img_channel)
+    
+    if max_val == 0
+        return img
+    end
+    
+    img_channel = img_channel./max_val
+    
+    img = colorview(RGB, img_channel, img_channel, img_channel)
+    img = Gray.(img)
```
johnnychen94, line 53: Why not just use `img_channel` in the following `imfilter`? This seems unnecessary.

docs/examples/sujoy_edge/sujoy_edge_demo.jl
```diff
@@ -0,0 +1,80 @@
...
+    t = sqrt(mean(grad))
+    bool_img = falses(size(grad))
+    bool_img[findlocalmaxima(grad)] .= true
+    return bool_img.*(sqrt.(grad) .> t)
+end
+
+input_img = testimage("cameraman")
+grayImg = sujoy(input_img,0,true)
+mosaicview(input_img,grayImg; nrow = 1)
+save("assets/sedge.png", grayImg) #src
```
johnnychen94, line 80: To follow the general Julia naming convention, I prefer to use `gray_img` instead of `grayImg`.

```suggestion
gray_img = testimage("cameraman")
gray_img_edge = sujoy(gray_img, 0, true)
mosaiview(gray_img, gray_img_edge; nrow = 1)

save("assets"sedge.png", gray_img_edge) #src
```

docs/examples/sujoy_edge/sujoy_edge_demo.jl
```diff
@@ -0,0 +1,80 @@
...
+# cover: assets/sedge.png
+# title: Sujoy Edge Demo
+# ---
+
+using Images,Statistics,ImageView,TestImages
+
+# Sujoy edge function
+
+
+# [Paper Link] (https://www.ijert.org/research/a-better-first-derivative-approach-for-edge-detection-IJERTV2IS110616.pdf)
```
johnnychen94, line 11: Would you mind to give some insightful explanation here on what Sujoy function is about. I mean, some description about it. It could a paragraph, or just one or two sentences.

johnnychen94 (commented):
I've made some comments to make it a better demo.

Also, could you move the file to `docs/examples/contours` folder? It doesn't look good to me to have a separate section for this demo.

The result actually looks quite good to me. But I'd like to get @zygmuntszpak involved here to see if he has some other comments.

zygmuntszpak:
I think Johnny has made some excellent suggestions. A further small detail:

```julia
edges = sujoy(img, kernel_nhood=0, just_filter=False)
```

Change  `False` to `false`. 

Also, since `kernel_nhood` toggles between a 4 or 8 connected neighbourhood, perhaps it is better to switch to a Bool type and instead of 
```julia
kernel_nhood : 0(default) or 1
```

have something along the lines of
```julia
four_connectivity : true (default) or false (uses 8-connectivity instead)
```

docs/examples/sujoy_edge/sujoy_edge_demo.jl
amarbit: Fix this

docs/examples/sujoy_edge/sujoy_edge_demo.jl
amarbit: Fix this

docs/examples/sujoy_edge/sujoy_edge_demo.jl
amarbit: Fix this

docs/examples/sujoy_edge/sujoy_edge_demo.jl
amarbit: Fix this

docs/examples/sujoy_edge/sujoy_edge_demo.jl
amarbit: Fix this

amarbit:
Hi 
did all changes as you suggested. Please accept it.

docs/examples/sujoy_edge/sujoy_edge_demo.jl
amarbit: Fix this

docs/examples/sujoy_edge/sujoy_edge_demo.jl
amarbit: Fix this

johnnychen94:
I've made some changes based on your codes, hope that's okay with you.

PR #165, opened 2020-11-21
Demo Addition-HistogramMatching

ashwani-rathee:
I think the issue of the cover image and that of the left part of the histogram still remains and more will be there after review.

### DIFF
--- a/docs/examples/spatial_transformation/histogram_matching.jl
+++ b/docs/examples/spatial_transformation/histogram_matching.jl
@@ -1,0 +1,35 @@
+# ---
+# cover: assets/woman_darkhair.tif
+# title: Histogram Matching
+# author: Ashwani Rathee
+# date: 2020-11-21
+# ---
+
+# This demo demonstrates the feature of histogram matching.
+
+# In histogram matching,the pixels of source image is manipulated so that the histogram 
+# of reference image matches with source image's histogram.
+
+# Here,histogram matching is shown for gray scale image.But,it can also be applied on 
+# RGB images too.Histogram matching can be used to normalize two images, when the images 
+# were acquired at the same local illumination (such as shadows) over the same location,
+# but by different sensors,atmospheric conditions or global illumination. 
+
+using Images,ImageCore,TestImages,ImageContrastAdjustment,ImageView
+img_source=testimage("woman_darkhair")
+img_reference=testimage("lena_gray_512")
+img_transformed=adjust_histogram(img_source, Matching(targetimg = img_reference))
+mosaicview(img_source,img_reference,img_transformed;nrow=1)
+
+# The images shown above are sourceimage,referenceimage,transformedimage.
+
+# To illustrate the effect of the histogram matching, we plot for source's histogram,
+# reference's histogram and the cumulative histogram. Clearly, the matched image
+# has the same cumulative histogram as the reference image for each channel.
+
+using Plots
+hist_source=histogram(vec(reinterpret(UInt8, img_source)));
+hist_reference=histogram(vec(reinterpret(UInt8, img_reference)));
+hist_transformed=histogram(vec(reinterpret(UInt8, img_transformed)));
+plot(hist_source,hist_reference,hist_transformed, layout =(1,3),size=(1000,333),legend=false,title=["source" "reference" "histograms matched"])
+


docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,35 @@
...
+mosaicview(img_source,img_reference,img_transformed;nrow=1)
+
+# The images shown above are sourceimage,referenceimage,transformedimage.
+
+# To illustrate the effect of the histogram matching, we plot for source's histogram,
+# reference's histogram and the cumulative histogram. Clearly, the matched image
+# has the same cumulative histogram as the reference image for each channel.
+
+using Plots
+hist_source=histogram(vec(reinterpret(UInt8, img_source)));
```
johnnychen94, line 31: In most cases, there are better tools than `reinterpret`.

```suggestion
hist_source = histogram(vec(channelview(img_source)));
```

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,35 @@
...
+# of reference image matches with source image's histogram.
+
+# Here,histogram matching is shown for gray scale image.But,it can also be applied on 
+# RGB images too.Histogram matching can be used to normalize two images, when the images 
+# were acquired at the same local illumination (such as shadows) over the same location,
+# but by different sensors,atmospheric conditions or global illumination. 
+
+using Images,ImageCore,TestImages,ImageContrastAdjustment,ImageView
+img_source=testimage("woman_darkhair")
+img_reference=testimage("lena_gray_512")
```
johnnychen94, line 20: Following the discussion in https://github.com/JuliaImages/TestImages.jl/issues/72, we don't use lena image in examples.

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,35 @@
...
+
+# In histogram matching,the pixels of source image is manipulated so that the histogram 
+# of reference image matches with source image's histogram.
+
+# Here,histogram matching is shown for gray scale image.But,it can also be applied on 
+# RGB images too.Histogram matching can be used to normalize two images, when the images 
+# were acquired at the same local illumination (such as shadows) over the same location,
+# but by different sensors,atmospheric conditions or global illumination. 
+
+using Images,ImageCore,TestImages,ImageContrastAdjustment,ImageView
```
johnnychen94, line 18: `ImageView` and `ImageCore` are not needed here

johnnychen94 (commented):
I'm not very sure the current content of this demo is useful because it doesn't give more information than what `?Matching` gives. Perhaps it would be more helpful to show how histogram matching can be used in other tasks. I'll leave it to @zygmuntszpak for a decision if he's available.

When I write a demo, I generally follow the following order and contents:

- summary
- background: what it is and how it can be useful
- how to do it
- (recommended) why it is done so
- (recommended) any references can be helpful

Codes and words need to be formatted accordingly. For example

```diff
- mosaicview(img_source,img_reference,img_transformed;nrow=1)
+ mosaicview(img_source, img_reference, img_transformed; nrow=1)

- # In histogram matching,the pixels of
+ # In histogram matching, the pixels of
```

docs/examples/spatial_transformation/histogram_matching.jl
ashwani-rathee: Yeah,I already saw that one, but unfortunately, I couldn't decide on which one to use(there is a need for more test images in my opinion)...Which ones should I use?? ..I did try for outside RGB images which looked pretty good as examples(cat-source,cafe-reference,transformed one)

ashwani-rathee:
I think what I should do is do it for RGB images and showing different histogram matching for each channel...and also an example of using build_histogram

johnnychen94:
Yes, it seems that RGB images are more illustrative (easier to identify the changes).

I don't know much about the application of histogram matching so I googled it and found the [skimage example](https://scikit-image.org/docs/dev/auto_examples/color_exposure/plot_histogram_matching.html) pretty good to copy (We may need a citation to it because they're the original)

According to https://github.com/scikit-image/scikit-image/blob/d254263605c6c173c563258d51f44e65782d4365/skimage/data/__init__.py#L896-L934, these two images are released under CC0 license so it's free to use. We can add these two images to TestImages and then use them here. See https://testimages.juliaimages.org/ for the guideline on how to add new images.

ashwani-rathee:
This can be closed now,since I can't find its source and so wasn't able to implement changes from its base

PR #166, opened 2020-11-22
add an introduction demo for indexed image

### LINKED ISSUE #132: [RFC] deprecate ColorizedArray and IndirectArray
This might not be a good suggestion but I'd like to propose it to get your thoughts

I'm wondering if there's a real-world case that these two lazy arrays make a performance difference; otherwise, these two array types introduce unnecessary abstraction and become a burden for both users and maintainers.

```julia
I = repeat([0.1, 0.2, 0.3, 0.4, 0.5], 1, 5)
idx = [1 2 3 4 5; 2 3 4 5 1; 3 4 5 1 2; 4 5 1 2 3; 5 1 2 3 4]
palette = [RGB(0.0, 0.0, 0.0), RGB(1.0, 0.0, 0.0), RGB(0.0, 1.0, 0.0), RGB(0.0, 0.0, 1.0), RGB(1.0, 1.0, 1.0)]

IndirectArray(idx, palette) == palette[idx] # true
IndirectArray(idx, palette)[i,j,...] == palette[idx[i,j,...]]
ColorizedArray(I, IndirectArray(idx, palette)) == I .* palette[idx] # true
```

```julia
mosaicview(palatte[img], I .* palatte[img]; nrow=1)
```
image

Reference: https://docs.julialang.org/en/v1/manual/arrays/#man-supported-index-types-1

> Supported index types: Any custom array of scalar indices that is a subtype of AbstractArray

@Drvi I find you're using `IndirectArray` in https://github.com/JuliaIO/PNGFiles.jl/pull/11, perhaps you have a better understanding.

johnnychen94:
We don't have an "image format" section, so I put this in the "Color Channel" section.

preview: https://juliaimages.org/previews/PR166/democards/examples/


docs/examples/color_channels/indexed_image.jl
```diff
@@ -0,0 +1,86 @@
...
+
+# This is doable because it follows the Julia [indexing
+# rules](https://docs.julialang.org/en/v1/manual/arrays/#man-supported-index-types) that `indices`
+# is an array of scalar indices. For example, the following equivalence holds:
+
+palatte[[1, 2]] == [palatte[1], palatte[2]]
+palatte[[1 2; 2 1]] == [palatte[1] palatte[2]; palatte[2] palatte[1]]
+#md nothing #hide
+
+# Now, let's do some math on it. For this example, we are storing the data using type `Float64`,
```
timholy, line 43:
```suggestion
# Now, let's analyze the resource usage of these two approaches. For this example, we are storing the data using type `Float64`,
```

docs/examples/color_channels/indexed_image.jl
```diff
@@ -0,0 +1,86 @@
...
+# where each `Float64` number requires 8 bytes (1 byte=8bits). Hence the normal representation
+# format requires `5*5*3*8 = 600` bytes storage in total. As a comparison, we only need `5*5*8 +
+# 5*3*8 = 320` bytes storage if we use the indexed image format. This is why indexed image requires
+# less memories.
```
timholy, line 47:
```suggestion
# 5*3*8 = 320` bytes storage if we use the indexed image format. This is why images with few distinct values can often be encoded more compactly as an indexed image.
```

docs/examples/color_channels/indexed_image.jl
```diff
@@ -0,0 +1,86 @@
...
+#
+# Although it does compress the data in this example, in real world applications, it can be quite
+# complicated and unclear whether you should or should not use indexed image format. There are two
```
timholy, line 50:
```suggestion
# Although it does compress the data in this example, in real world applications, it is not always
# clear whether you should or should not use indexed image format. There are two
```

docs/examples/color_channels/indexed_image.jl
```diff
@@ -0,0 +1,86 @@
...
+# main drawbacks of it:
+#
+# - indexed images can require more memories if `length(unique(img))` is too large.
```
timholy, line 53:
```suggestion
# - indexed images can require more memory if `length(unique(img))` is too large.
```

docs/examples/color_channels/indexed_image.jl
```diff
@@ -0,0 +1,86 @@
...
+# - indexing into an indexed image requires two `getindex` operations, so using it can be
+#   theoratically slower than a normal image representation. This is a typical [space-time
```
timholy, line 55:
```suggestion
#   slower than a direct image representation (and not amenable to SIMD vectorization). This is a typical [space-time
```

timholy (commented):
Looks good, thanks for adding this!

johnnychen94:
All great suggestions! It definitely improves readability.

PR #168, opened 2020-11-25
introduce ColorBlendModes in AlphaCompositing demo

johnnychen94:
@kimikage This demo was written during the same time when you're developing ColorBlendModes.jl. For this demo, I only intend to introduce how PaddedViews and OffsetArrays are used; and the color composition is really only one typical example of it. So yeah, I think it's better to introduce your wonderful package here.


kimikage:
Thank you for the introduction.

As you have mentioned, this demo is focused on PaddedViews and OffsetArrays and categorized under "Spatial transformation". That allows ColorBlendModes.jl (and its documentation) to focus only on pixel-level operations. :smiley:

However, the title of this demo, "Alpha Compositing," is somewhat misleading. Many people hear the term alpha compositing and think of the alpha blending or the "over" operation *with opacity*.

So it might be a good idea to retitle this demo "Layer Compositing".

PR #169, opened 2020-12-05
Update arrays_colors.md

krishna-vasudev:
Helps in understanding the concept better. It gives  clear understanding of how julia treats when it encounters integers.

### DIFF
--- a/docs/src/tutorials/arrays_colors.md
+++ b/docs/src/tutorials/arrays_colors.md
@@ -292,8 +292,8 @@ go to compare images: in most number systems we would agree that `255
 careful when converting from one representation to another.
 Conversely, using these Julia packages **there is no discrepancy in
 "meaning" between the encoding of images represented as floating point
-or 8-bit (or 16-bit) fixed-point numbers: 0 always means "black" and 1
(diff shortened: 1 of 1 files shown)


johnnychen94:
Technically, Integers are not the first-class citizen in JuliaImages as we have `N0f8` things (defined in FixedPointNumbers.jl) to mimics the integer values.

```julia
julia> using ImageCore

julia> reinterpret(N0f8, UInt8(255))
1.0N0f8
```

It's designed so that we don't get confused about what `1` actually means. And this is why we can safely define `1` as white.

krishna-vasudev:
> Technically, Integers are not the first-class citizen in JuliaImages as we have `N0f8` things (defined in FixedPointNumbers.jl) to mimics the integer values.
> 
> ```julia
> julia> using ImageCore
> 
> julia> reinterpret(N0f8, UInt8(255))
> 1.0N0f8
> ```
> 
> It's designed so that we don't get confused about what `1` actually means. And this is why we can safely define `1` as white.

Thanks, that is something beautiful about juliaimages.

PR #171, opened 2020-12-06
correct colorview result in conversions_views

krishna-vasudev:
There was a mistake in the output. So ,I corrected it.


johnnychen94:
A suggestion for your future PR: it might be good to have an informative title compared to `Update xxx.md`.

krishna-vasudev:
> A suggestion for your future PR: it might be good to have an informative title compared to `Update xxx.md`.

I will take care of that. As i am new to opensource, i was unaware of that. Thanks for the information.

PR #172, opened 2020-12-08
remove extra space before the three-backticks

szcf-weiya:
there is an extra space before the ending three-backticks of the code block, which causes an abnormal display in the webpage
image


johnnychen94:
thanks!

PR #175, opened 2020-12-13
addition_of_example_in_imageaxes

krishna-vasudev:
Helps in understanding the orientation of axes better.


johnnychen94:
Hmmm, not sure if I understand your intention here. Could you explain a little bit more what problem you're trying to address?

krishna-vasudev:
> Hmmm, not sure if I understand your intention here. Could you explain a little bit more what problem you're trying to address?

Suppose the example in docs, contains an example for z-axis representation, so i have added one for x-axis and y-axis, so the reader gets a clear imagination of these axes too. Although, they can run it themselves and see, but adding it in the docs makes it more rich.

johnnychen94:
Alright, adding more results could be helpful for readers, so I'll merge this.

Issue #164, opened 2020-10-28
Discussion on making more demos to  showcase abilities of Julia and libraries for Images


ashwani-rathee:
Hey Guys,I am Ashwani Rathee,undergraduate cs student at UIET,Panjab University,India
Saw this on: https://julialang.org/jsoc/gsoc/images/
I have explored JuliaImages in past: https://github.com/ashwani-rathee/Learning-Julia/blob/master/ImageProcessing%20in%20Julia.ipynb
And would love to contribute through examples: @timholy and @zygmuntszpak guide me on this and give me some direction to work on:)
```
Wide-ranging demos
For new or occasional users, JuliaImages would benefit from a large collection of complete worked examples organized by topic. While the current documentation contains many "mini-demos," they are scattered; an organized page would help users quickly find what they need. We have set up a landing page, but many more demos are needed. Scikit-image is one potential model.

This "project" might also be split among multiple students who contribute demos as part of their work in a focused area of JuliaImages.
```

johnnychen94:
Hi Ashwani, thanks for your interest in JuliaImages!

Currently, demos in JuliaImages are managed by [DemoCards](https://github.com/johnnychen94/DemoCards.jl). Existing demos are spreaded in multiple packages: [juliaimages.github.io](https://github.com/JuliaImages/juliaimages.github.io) as the main doc, [ImageFiltering.jl/docs](https://github.com/JuliaImages/ImageFiltering.jl) for image filtering operations. The demos are written using a mixed [Documenter markdown syntax](https://juliadocs.github.io/Documenter.jl/v0.25/man/syntax/), [Literate julia syntax](https://fredrikekre.github.io/Literate.jl/v2.7/fileformat/) and [DemoCards file syntax](https://johnnychen94.github.io/DemoCards.jl/v0.3/democards/quickstart/). You don't need to master all the syntax in the first go, as there are now many examples available for you to copy and modify. #144 is a good example of a demo PR.

We don't have a specific wish list of what demos should be added, and it can be potentially dangerous for us to give a list as we're now "advanced" developers; things that are trivial to us might not be so trivial to others. For this reason, you can start with something simple based on your own decision and understanding.

ashwani-rathee:
Hey @johnnychen94 ,for the demos..is there any kind of preference for mosaicview over the plots.jl for showing the images??

johnnychen94:
[...]

PR #178, opened 2021-02-01
Histogram Matching demo

ashwani-rathee:
Demo for histogram matching has been added with this #164

Needs review

closes #165


docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,77 @@
...
+# cover: assets/chelsea.png
+# title: Histogram Matching
+# author: Ashwani Rathee
+# date: 2020-11-21
+# ---
+
+# This demo demonstrates the feature of histogram matching.
+
+# In histogram matching,the pixels of source image is manipulated so that the histogram 
+# of reference image matches with source image's histogram.
```
johnnychen94, line 11:
```suggestion
During histogram matching, the pixels of the source image is manipulated so that the histogram of it matches with that of the reference image.
```

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,77 @@
...
+
+# Here,histogram matching is shown for gray scale image.But,it can also be applied on 
+# RGB images too.Histogram matching can be used to normalize two images, when the images 
+# were acquired at the same local illumination (such as shadows) over the same location,
+# but by different sensors,atmospheric conditions or global illumination.
```
johnnychen94, line 16:
```suggestion
# Here, histogram matching is shown for the gray scale image. But, it can also be applied on 
# RGB images too. Histogram matching can be used to normalize two images where the images 
# are acquired at the same local illumination (such as shadows) over the same location
# but by different sensors, atmospheric conditions, or global illumination. 
```

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,77 @@
...
+img_source = testimage("chelsea");
+img_reference = testimage("coffee");
+
+# Applying histogram matching on source image using reference image
+
+img_transformed=adjust_histogram(img_source, Matching(targetimg = img_reference));
+
+# **Results** : 
+
+mosaicview(img_source,img_reference,img_transformed;nrow=1)
```
johnnychen94, line 36:
```suggestion
mosaicview(img_source, img_reference, img_transformed; nrow=1)
```

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,77 @@
...
+
+# From Left to Right: Source Image,Reference Image,Transformed Image
```
johnnychen94, line 38:
```suggestion
# From Left to Right: Source Image, Reference Image, Transformed Image
```

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,77 @@
...
+
+# We want to treat our RGB image as 3-component vector,so that we 
+# can access data from red,green,channels to show
+# histogram matching effect.
```
johnnychen94, line 42:
```suggestion
# We want to treat our RGB image as a 3-component vector so that we 
# can access data from red green and blue channels to show
# histogram matching effect.
```

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,77 @@
...
+
+img_source=channelview( img_source);
+img_reference=channelview( img_reference);
+img_transformed=channelview( img_transformed);
```
johnnychen94, line 46:
```suggestion
img_source = channelview(img_source);
img_reference = channelview(img_reference);
img_transformed = channelview(img_transformed);
```

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,77 @@
...
+
+# Clearly, the matched image has the same cumulative histogram as the 
+# reference image for each channel.
+
+# Analysing change of data in color channel red
+
+hist_sred=histogram(vec(img_source[1,:,:]));
+hist_rred=histogram(vec(img_reference[1,:,:]));
+hist_tred=histogram(vec(img_transformed[1,:,:]));
+plot(hist_sred,hist_rred,hist_tred, layout =(1,3),size=(1000,333),legend=false,title=["source" "reference" "histograms matched"])
```
johnnychen94, line 58:
```suggestion
hist_sred = histogram(vec(img_source[1,:,:]));
hist_rred = histogram(vec(img_reference[1,:,:]));
hist_tred = histogram(vec(img_transformed[1,:,:]));
plot(hist_sred, hist_rred, hist_tred;
    layout=(1,3),
    size=(1000,333),
    legend=false,
    title=["source" "reference" "histograms matched"])
```

johnnychen94: It might be clearer to run a for loop thing over R,G,B channels and make a 3*3 plot on it.

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,77 @@
...
+# Analysing change of data in color channel blue
+
+hist_sblue=histogram(vec(img_source[3,:,:]));
+hist_rblue=histogram(vec(img_reference[3,:,:]));
+hist_tblue=histogram(vec(img_transformed[3,:,:]));
+plot(hist_sblue,hist_rblue,hist_tblue, layout =(1,3),size=(1000,333),legend=false,title=["source" "reference" "histograms matched"])
+
+# ### References:
+# 1. Python Example Implementation of Histogram Matching: https://scikit-image.org/docs/stable/auto_examples/color_exposure/plot_histogram_matching.html#sphx-glr-auto-examples-color-exposure-plot-histogram-matching-py
+
```
johnnychen94, line 76: The cover image is not created here: `assets/chelsea.png`.

I also noticed that one of your [previous demo](https://juliaimages.org/stable/democards/examples/spatial_transformation/SpatialTransformations/#Cropping,-Resizing-and-Rescaling) forgets to do this, could you add it up?

johnnychen94 (commented):
`Plots` should be included in `Project.toml` with an appropriate compat.

ashwani-rathee:
> `Plots` should be included in `Project.toml` with an appropriate compat.

@johnnychen94 Hey,how can i do that?
The last time I did something similiar,I used below command for project.toml generation
"Pkg.generate("MusicProcessing")" [#5]([https://github.com/JuliaMusic/MusicProcessing.jl/pull/5)

johnnychen94:
You could do it by activating the current project and adding it:

```bash
$ julia --project=.

pkg> add Plots
```

and then manually modifying `Project.toml` to add a compat entry for `Plots`( `Plots = [1]`, maybe).

ashwani-rathee:
little confused about where to add the chelsea.png in src/assets? or in democards/examples/covers

johnnychen94:
In a relative directory `assets/chelsea.png` and it's generated in runtime. You can take a look at other examples, e.g., https://github.com/JuliaImages/juliaimages.github.io/blob/18356b56e102c4b7d97cfbf3f9947536244a1c4a/docs/examples/image_quality_and_benchmarks/structural_similarity_index.jl#L46

ashwani-rathee:
This adds plots and cover properly,should I combine the graphs together?

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,78 @@
...
+img_reference = testimage("coffee");
+
+# Applying histogram matching on source image using reference image
+
+img_transformed=adjust_histogram(img_source, Matching(targetimg = img_reference));
+
+# **Results** : 
+
+mosaicview(img_source,img_reference,img_transformed;nrow=1)
+save("assets/chelsea.png",img_source) #src
```
johnnychen94, line 37: I hope we can keep some more meaningful results as our cover here so that potential readers immediately get what "Histogram Matching" is. Perhaps a gif that shows the matching progress would be good enough?

johnnychen94 (commented):
Could you format the codes properly using [JuliaFormatter.jl](https://github.com/domluna/JuliaFormatter.jl), vscode, or atom? It would be better if readers get a consistent reading experience through all the docs and demos.

I'd also recommend [Grammarly](https://www.grammarly.com/) to format/improve the languages you used here.

docs/examples/spatial_transformation/histogram_matching.jl
ashwani-rathee: Sure :)

ashwani-rathee: @johnnychen94 I have cleaned up the spaces and stops part of the demo with grammarly and formatted it with vs code.Just learned to create gif with save with help of @timholy from here :https://discourse.julialang.org/t/plotting-a-sequence-of-images-as-a-gif/23808/25
By creating a 3d array of images
```
using FileIO, ImageMagick, Colors, FixedPointNumbers
using TestImages
img1=testimage("cameraman")
img2=testimage("fabio_color_512")
img3=testimage("lena_color_512")
save("/home/ashwani/test.gif", cat(img1,img2,img3;dims=3))
```
But issue is that, whenever dimension mismatch happens it shows error....so I think what I could do is that I create a gif of initial chelsea and final chelsea as their size will remain same

johnnychen94: Maybe use `PaddedViews` to make the sizes equal:

```julia
cat(sym_paddedviews(colorant"white", img1,img2,img3)...; dims=3)
```

Also, it should be better to use `ImageMagick.save` explicitly since other image IO backends don't support `gif` saving yet. Although not documented, `ImageMagick.save` also accepts a `fps` keyword.

An live example for this can be found in https://github.com/JuliaImages/juliaimages.github.io/blob/4749786007ffd4553ca265bd6115126e3a1047b0/docs/examples/color_channels/color_separations_svd.jl#L51

johnnychen94:
Similar to #183, I do think this demo can be written in a more concise while still informative way. There will be many new users looking at the demo so I hope they can be polished into some kind of "standard" form. I'll wait for a few more days so that you get more time to think about it.

ashwani-rathee:
@johnnychen94 I did try to update by looking at canny rework and shortening the graph code by creating array,list,loop usage

johnnychen94:
Curious to ask, have you ever tried out the partial build functionality of DemoCards?

```julia
julia> using DemoCards

julia> preview_demos("docs/examples/spatial_transformation/histogram_matching.jl")
```

ashwani-rathee:
Actually,I have.It's pretty good at what it's supposed to do,though it did confused me as it directs to index.html and not the demo directly.

PR #177, opened 2021-02-01
Canny Edge Filter Demo

ashwani-rathee:
Demo for canny edge filter has been added with this #164

Needs review

PR #183, opened 2021-03-20
new demo: Canny operator (rework #177)

johnnychen94:
@ashwani-rathee I feel the previous PR #177 is less informative so I did a rewrite here, hope you don't mind.

I also noticed that words are copied almost directly from [the skimage example](https://scikit-image.org/docs/dev/auto_examples/edges/plot_canny.html), it's allowed but if we do this then we should properly add its [license](https://scikit-image.org/docs/dev/license.html) in the example. I wish we could keep things simple, so I just removed them in this PR. Even if we're doing an open-source project, license is still important so if there're similar issues in your future PR, please inform us; you can read and learn, but try to avoid a copy and modify.

preview: https://juliaimages.org/previews/PR183

closes #177

Co-authored-by: Ashwani Rathee <ab669522@gmail.com>


ashwani-rathee:
@johnnychen94 Okay,I'll avoid any copy pasting whatsoever of the words(I didn't knew that part also comes under copyright).

As for the demo and rewrite,I feel I did #177 fine but this would be better,since you have way more experience than me with specific details and using efficient methods/functions

johnnychen94:
Most people aren't aware of this at first. For me it's https://github.com/JuliaImages/ImageFiltering.jl/pull/81 😆 

For this demo, I feel the image generation is less important so I wrapped it with a function, I feel by doing this can emphasize that we're doing a demo on the canny operator. It's just my preference, though.

ashwani-rathee:
Ouch!! @johnnychen94,that was pretty rough,My first time with this copyright issue handling is far more considerate and digestible

PR #185, opened 2021-03-23
Histogram Matching demo (cont. #178)

johnnychen94:
@ashwani-rathee I can't push to your branch so again I created a new PR here.

In commit ae18004 there're two changes:

- language modification for accuracy and correctness.
- simplify the codes and try to make them self-explained.

We don't have a so-called standard for how demos or documentation are written, but there's a good [blog](https://documentation.divio.com/) talking about this issue and most of us just try to follow his suggestions. Also here's [his talk in youtube](https://www.youtube.com/watch?v=t4vKPhjcMZg).

preview: https://juliaimages.org/previews/PR185/examples/spatial_transformation/histogram_matching/#Histogram-Matching

closes #178


docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,67 @@
...
+# Applying histogram matching on source image using the reference image.
+# Here we use `adjust_histogram` function from ImageContrastAdjustment.jl.
+# It returns a histogram matched image with a granularity of `nbins`, i.e., number of bins. 
+# The first argument `img` is the image to be matched, and the second argument `targetimg` is 
+# the image with the desired histogram to be matched to.
+
+img_transformed = adjust_histogram(img_source, Matching(targetimg = img_reference))
+mosaicview(img_source, img_reference, img_transformed; nrow = 1)
+
+save("assets/chelsea.gif", cat(img_source, img_transformed; dims = 3); fps=2) #src
```
johnnychen94, line 40: Here I deliberately set a lower fps rate so readers can see the changes more clearly.

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,67 @@
...
+using ImageContrastAdjustment
+using Images
+using Plots
+using TestImages
+
+# Load example images: one source image and one reference image 
+
+# !!! note
+#     You need to use `julia >= v"1.3.0"` and `TestImages >= v"1.3.1"` in order to load these
+#     two test images.
```
johnnychen94, line 27: Instead of saying a "recent TestImages", saying the version is more accurate a way.

docs/examples/spatial_transformation/histogram_matching.jl
```diff
@@ -0,0 +1,67 @@
...
+mosaicview(img_source, img_reference, img_transformed; nrow = 1)
+
+save("assets/chelsea.gif", cat(img_source, img_transformed; dims = 3); fps=2) #src
+
+# To show the effect of histogram matching, we plot for each RGB channel the histogram.
+
+hist_final = [histogram(vec(c.(img)))
+    for c in (red, green, blue)
+    for img in [img_source, img_reference, img_transformed]
+]
```
johnnychen94, line 47: This is mostly equivalent to the for loop version in #178 but is shorter and thus (I believe) more readable.

docs/examples/spatial_transformation/histogram_matching.jl
ashwani-rathee: I do know about the fps change,but didn't feel it made a big effect.There is another issue still,you mentioned about the padding of gifs for images but still both of us forgot about that,the cat gif needs to be padded vertically to fit in the cover.

ashwani-rathee: Other changes make it more concise sure,should have included the testimage note for people who don't update on time(like me) 😄 .But isn't the loop too compact ?

johnnychen94: > There is another issue still,you mentioned about the padding of gifs for images but still both of us forgot about that,the cat gif needs to be padded vertically to fit in the cover.

Ah yeah, but then I thought excluding `img_reference` from the gif is also acceptable...

johnnychen94: > But isn't the loop too compact ?just saying :) It's more intuitive I know
Well.. yeah it's a matter of taste here but I kept trying to bring more things like this to encourage people to code in a functional way.. to show how expressive the Julia codes are...

PR #191, opened 2021-05-05
new demo: block-by-block visualization as GIF

johnnychen94:
The cover image:

block_visualization


docs/examples/visualization/block_visualization.jl
```diff
@@ -0,0 +1,64 @@
...
+# back the axes information for each block. We also use the original image as the
+# background.
+
+function create_frames(img, background, tile_indices; rf=0.8, rb=1-rf)
+    map(tile_indices) do R
+        ## add back axes information to given blcok
+        block = OffsetArray(img[R...], OffsetArrays.Origin(first.(R)))
+        canvas, block = paddedviews(zero(eltype(background)), background, block)
+        frame = clamp01!(@. rb*canvas .+ rf*block)
+    end
```
johnnychen94, line 46: There are extra memory allocations here so it's not very efficient. The main reason I didn't optimize it is that the current `ImageShow.gif` is a little bit broken when dealing with non-`Array` array types.

docs/examples/visualization/block_visualization.jl
```diff
@@ -0,0 +1,64 @@
...
+# author: Johnny Chen
+# date: 2021-05-05
+# ---
+
+using Images
+using PaddedViews, OffsetArrays
+using TestImages
+using ImageBinarization
+
+# !!! tip
```
johnnychen94, line 14: Maybe it's better to use `note` instead of `tip` here.

PR #193, opened 2021-05-24
Add demo:Gaussion Pyramid

parkma99:
This is  a very small demonstration about Gaussion Pyramid.This is also my first try in JuliaImages.I would like to display the result like this picture.

 pyramid

But I fail it .Thanks so much if reviewers can give me help!

### DIFF
--- a/docs/examples/spatial_transformation/gaussian_pyramid.jl
+++ b/docs/examples/spatial_transformation/gaussian_pyramid.jl
@@ -1,0 +1,28 @@
+# ---
+# cover: assets/lake_color.tif
+# title: Gaussian Pyramid
+# author: Park ma
+# date: 2021-5-24
+# ---
+
+# This demo illustrates the use of [Gaussian Pyramid](https://juliaimages.org/stable/function_reference/#Images.gaussian_pyramid)
+
+# Gaussian Pyramid is a technique in image processing that breaks down 
+# an image into successively smaller groups of pixels to blur it. 
+
+using TestImages,Images
+
+img = testimage("lake_color")
+
+# We will build image pyramids though the Gaussian Pyramid.
+
+img_pyramids = gaussian_pyramid(img,5,2,2.0)
+
+mosaicview(img_pyramids...;nrow=1,rowmajor=true)
+
+# ---save covers ---#src
+using FileIO #src
+save("assets/lake_color.tif",img)#src
+
+
+


docs/examples/spatial_transformation/gaussian_pyramid.jl
```diff
@@ -0,0 +1,28 @@
...
+
+# We will build image pyramids though the Gaussian Pyramid.
+
+img_pyramids = gaussian_pyramid(img,5,2,2.0)
+
+mosaicview(img_pyramids...;nrow=1,rowmajor=true)
+
+# ---save covers ---#src
+using FileIO #src
+save("assets/lake_color.tif",img)#src
```
johnnychen94, line 25: Can we use the output of `mosaicview` as the cover? That would be most visually attractive to me.

docs/examples/spatial_transformation/gaussian_pyramid.jl
```diff
@@ -0,0 +1,28 @@
...
+# Gaussian Pyramid is a technique in image processing that breaks down 
+# an image into successively smaller groups of pixels to blur it. 
+
+using TestImages,Images
+
+img = testimage("lake_color")
+
+# We will build image pyramids though the Gaussian Pyramid.
+
+img_pyramids = gaussian_pyramid(img,5,2,2.0)
```
johnnychen94, line 19: For self completeness, I feel that we need to explain each input of `guassian_pyramid` here, and the general idea of how to choose the parameters.

johnnychen94 (commented):
The first try (by @ashwani-rathee) and recent discussion on this can be found in https://github.com/JuliaImages/juliaimages.github.io/issues/164#issuecomment-733737540 I'll summarize it briefly here:

- `ImageTransformations.restrict` is a more efficient way to construct a pyramid (although not _Gaussian_ pyramid).
- need to extend `paddedviews` and `sym_paddedviews` to remove the blank area.

I would suggest you compare these two pyramid construction methods in the same demo.

For the second point, I'm fine to temporarily have a less beautiful version of it. After all, it's just about how we show the image.

The most important aspect that I would like to see in this demo is a small application that utilizes the pyramid, instead of simply building the pyramid, which is already made into one line. How do you think?

---

> But I fail it 

You can use `mosaicview` to tile the downsampled images along column direction, i.e., `ncol=1`, and then again use `mosaicview` to tile the original one with the output of previous `mosaicview`.

For example,

```julia
mosaicview(original,
    mosaicview(pyramid; ncol=1)
)
```

parkma99:
Thanks for your helping.
I will compare these two pyramid construction methods.
> The most important aspect that I would like to see in this demo is a small application that utilizes the pyramid, instead of simply building the pyramid, which is already made into one line. How do you think?

After googled the pyramid,I would like to make a demonstration about image blending using pyramid.
It needs upsample. I am not sure that the function `imresize`  can be used for upsample.

johnnychen94:
>  I am not sure that the function imresize can be used for upsample.

`imresize` supports different interpolation methods now (e.g., `method= Lanczos4OpenCV()`, you can take a try and see if there are actual issues. If there were we can discuss and figure out a way to fix or work around that.

parkma99:
> After googled the pyramid,I would like to make a demonstration about image blending using pyramid.

But I don't know how to mask the image in Julia. Do somebody know how to do it?

johnnychen94:
> But I don't know how to mask the image in Julia. Do somebody know how to do it?

Something like https://juliaimages.org/stable/democards/examples/spatial_transformation/layer_compositing/#Layer-Compositing and https://github.com/kimikage/ColorBlendModes.jl?

PR #201, opened 2021-07-02
Image Binarization

ashwanirathee:
- [x] Supposed to provide demonstrations and examples for ImageBinarization.jl

Status : `Early Stage`, `Not Ready`

Preview Link : 

Knowingly/unknowingly, current version does show http related methods and metaprogramming support usage here which might be a good thing.

Notes: 
- To binarize a brain scan(.nii)(3d image), needs to be figured out
- ImageMagick.load loses a lot of detail this way of url download
 ```julia
using HTTP, ImageMagick

function get_image(url)
     response = HTTP.get(url)
     buffer = IOBuffer(response.body)
     img = ImageMagick.load(buffer)
end
# This returns a very bad image for some reason
```

Better method to download images
```julia
using Images
function get_image(url)
    img = mktemp() do fn, f
        download(url, fn)
        load(fn)
    end
    img_resized = imresize(img, ratio = 1 / 2)
end

img_coins = get_image("https://i.imgur.com/91IOKtJ.png")
```


docs/examples/image_binarization/image_binarization.jl
```diff
@@ -0,0 +1,128 @@
...
+]
+function binarize_methods(img_input, algs)
+    imgs_binarized = Array[]
+    for i in algs
+        alg = getfield(ImageBinarization, Symbol(i))()
+        img_input1 = binarize(img_input, alg)
+        push!(imgs_binarized, img_input1)
+    end
+    return imgs_binarized
+end
```
johnnychen94, line 71: What's the advantage of this compared to

```julia
algs = [AdaptiveThreshold(), Balanced(), Entropy(), Intermodes(),
        MinimumError(), Moments(), Niblack(), Otsu(),
        Polysegment(), Sauvola(), UnimodalRosin(), Yen()]
outs = map(alg->binarize(img, alg), algs)
```
?

docs/examples/image_binarization/image_binarization.jl
```diff
@@ -0,0 +1,128 @@
...
+img = Gray.(img)
+
+# Let's first rotate our image
+
+trfm = recenter(RotMatrix(pi / 30), center(img));
+imgw = warp(img, trfm)
+
+# Now let's zoom in a bit
+
+imgw = parent(imgw)[65:490, 65:490]
```
johnnychen94, line 37: For this, we should instead introduce `imrotate(img, degree, indices)` API

docs/examples/image_binarization/image_binarization.jl
```diff
@@ -0,0 +1,128 @@
...
+using Images, ImageBinarization, TestImages
+using  CoordinateTransformations, Rotations, ImageTransformations
+
+# Original Image
+
+img = testimage("sudoku")
+
+# Grayscale version as binarize! only accepts Grayscale images
+
+img = Gray.(img)
```
johnnychen94, line 28: I believe it is a mistake that `binarize` doesn't support `RGBA` images; it doesn't support `RGB` images. We should open an issue in ImageBinarization and provide a PR for it.

I checked the `alpha.(img)` for the sudoku image and then I realize it's all 1: `all(x->x==1, alpha.(img)) == true`. Since the alpha channel is not used at all, I think we should update our testimage and using the RGB format.

docs/examples/image_binarization/image_binarization.jl
```diff
@@ -0,0 +1,128 @@
...
+# | MinimumError |  Moments | Niblack |
+# | ![](assets/test5.png) | ![](assets/test6.png) | ![](assets/test7.png) |
+# | Polysegment | Sauvola | UnimodalRosin |
+# | ![](assets/test9.png) | ![](assets/test10.png) | ![](assets/test11.png) |
+# | Intermodes | Otsu | Yen |
+# | ![](assets/test4.png) | ![](assets/test8.png) | ![](assets/test12.png) |
+
+# We can choose one of methods based on the results here, and use `OCReact.jl` 
+# which is based on Tesseract OCR to find the content in the sudoku and then
+# solve the sudoku.
```
johnnychen94, line 93: I can't find `OCReact.jl` in JuliaHub and GitHub.

docs/examples/image_binarization/image_binarization.jl
```diff
@@ -0,0 +1,128 @@
+# ---
+# title: Image Binarization
+# author: Ashwani Rathee
+# date: 2021-7-2
+# ---
```
johnnychen94, line 5: cover image is missing here.

johnnychen94 (commented):
I was expecting this demo to provide more information than https://zygmuntszpak.github.io/ImageBinarization.jl/dev/. But it turns out that it only
changes the input image?

The idea of solving sudoku sounds a great case to me but I didn't find the package OCReact

docs/examples/image_binarization/image_binarization.jl
ashwani-rathee: https://github.com/leferrad/OCReract.jl extra r issue

docs/examples/image_binarization/image_binarization.jl
ashwani-rathee: I didn't know we could make array of functions like this. Suggested version would be more concise then..

ashwani-rathee:
Actually, there isn't much to add with what I know. I think we can add the build histogram(with its graph) and show the threshold..

docs/examples/image_binarization/image_binarization.jl
ashwani-rathee: Sure, we need to add imrotate example here too:https://juliaimages.org/latest/pkgs/transformations/

johnnychen94: Actually, it's already supported in ImageTransformations, e.g., `imrotate(img, 0.3, axes(img))`.a

The transformation documentation should be removed and redirected to our new one: https://juliaimages.org/ImageTransformations.jl/dev

johnnychen94:
I'm a little bit redundant to the current version because it adds basically nothing new compared to the existing index page of ImageBinarizations. The API itself is quite simple..

I feel we need some application that uses ImageBinarization as a part of its pipeline. The sudoku solver sounds interesting, fingerprint enhancement together with ImageMorphology seems also a good candidate. How do you think?

docs/examples/image_binarization/image_binarization.jl
ashwani-rathee: I think we can use this as an opportunity to discuss a little bit about alpha channel. I didn't even know it was a thing, thanks to images.jl handling it all in background. Will sudoku image throw errors if sudoku is used in any other place where RGB is required?

docs/examples/image_binarization/image_binarization.jl
johnnychen94: I think we should use https://github.com/pixel27/Tesseract.jl instead because it's backed by Tesseract_jll

docs/examples/image_binarization/image_binarization.jl
johnnychen94: > Will sudoku image throw errors if sudoku is used in any other place where RGB is required?

Can you explain this a bit more? `RGBA` is not universally supported by the entire JuliaImages stack so we might just directly update our soduku image with `RGB` format in TestImages.

docs/examples/image_binarization/image_binarization.jl
ashwani-rathee: OCReract.jl is better maintained I think, no updates in Tesseract.jl in last 15 months. But https://github.com/JuliaPackaging/Yggdrasil shows that Tesseract folder was updated in past 12 days. I don't know enough binary builds, so we can go for Tesseract.jl if you think that's a better option

johnnychen94: > OCReract.jl is better maintained I think

Nope, `OCReract.jl` requires the user to manually install Tesseract and make it available in PATH. While Tesseract.jl ships a prebuilt Tesseract using BinaryBuilder so that user just need `pkg> add Tesseract` and that's all.

PR #203, opened 2021-07-04
Image Morphology 2: Convexhull, Filling,Thinning, Clearborder

ashwani-rathee:
- [x] Covers  operations from ImageMorphology.jl not covered in the other demo

There are some other functions hiding in there that I don't know use of yet, so need to figure that out.

Status: `Early Stage`, `Not Ready`
Preview Link: [Link](https://juliaimages.org/previews/PR203/examples/image_morphology/morphological_operations/#Convex-Hull,-Filling,-Thinning,-ClearBorder)


docs/examples/image_morphology/morphological_operations.jl
```diff
@@ -0,0 +1,73 @@
...
+
+# Thresholding of image
+
+edges1, counts1 = build_histogram(Gray.(img_src), 256)
+t1 = find_threshold(UnimodalRosin(), counts1[1:end], edges1)
+img_input = zeros(Bool, axes(Gray.(img_src)))
+for i in CartesianIndices(img_src)
+    img_input[i] = Gray.(img_src[i]) < t1 ? 0 : 1
+end
+img_input
```
johnnychen94, line 25: Isn't this `binarize(Gray{Bool}, img_src, UnimodalRosin())`?

It currently doesn't work, so `Gray{Bool}.(binarize(img_src, UnimodalRosin()))` (https://github.com/zygmuntszpak/ImageBinarization.jl/issues/83)

docs/examples/image_morphology/morphological_operations.jl
```diff
@@ -0,0 +1,73 @@
...
+# ## Convex Hull
+
+# Convex Hull operation outer most cover-like boundary of a binary image and 
+# returns the vertices of convex hull as a CartesianIndex array.
+
+cordinates = convexhull(img_input)
+push!(cordinates, cordinates[1])
+img_convex_hull = zeros(RGB, axes(Gray.(img_src)))
+draw!(img_convex_hull, Path(cordinates), RGB(1))
+img_convex_hull
```
johnnychen94, line 41: Maybe we can draw the convex hull on top of the original image?

johnnychen94 (commented):
I'm not a morphology expert, I would expect to know more information when reading this example and its output, but unfortunately, I am still confused here...

From my own reading experience: the clear border and image fill seems to lack enough explanations on the operation and output.

docs/examples/image_morphology/morphological_operations.jl
ashwani-rathee: neither one of those work really

johnnychen94: Because `morphology_test_512` is `RGBA`, so need to change it to: `RGB.(testimage("morphology_test_512"))`.

The better approach is to update the upstream image in TestImages.

ashwani-rathee: More like `binarize(Gray.(img_src), UnimodalRosin()) .> 0.5`
Operations require 512*512 BitMatrix, and not Gray.{Bool}, slight modification works

docs/examples/image_morphology/morphological_operations.jl
ashwani-rathee: sure thing 👍🏻

docs/examples/image_morphology/morphological_operations.jl
ashwani-rathee: I would need to update both of those 😓

PR #159, opened 2020-09-23
[WIP] build pkgs documentation with DemoCards

johnnychen94:
This is still a WIP PR; it requires the master branch of DemoCards.

The idea is to move `docs/src/pkgs` to `docs/pkgs` and build it with DemoCards, this enables us to collect remote documentation in sub packages (e.g., ImageFiltering), and furthermore, we could gradually move the current pkgs docs back to their repo.

This is the new folder structure, note that `docs/pkgs/config.json` is added for order specification:

```text
docs
├── examples
│   ├── color_channels
│   ├── config.json
│   ├── contours
│   ├── image_quality_and_benchmarks
│   ├── index.md
│   └── spatial_transformation
├── make.jl
├── pkgs
│   ├── axes
│   ├── config.json
│   ├── features
│   ├── introduction
│   ├── metadata
│   ├── segmentation
│   └── transformations
└── src
    ├── api_comparison.md
    ├── assets
    ├── function_reference.md
    ├── index.md
    ├── install.md
    └── tutorials
```

PR #205, opened 2021-07-12
Transfer of BRIEF,ORB,BRISK,FREAK, HOG

ashwani-rathee:
- [x] Transfers ImageFeatures.jl algorithms tutorials here except GLCM,Local binary patterns
Preview Link: https://juliaimages.org/previews/PR205/examples/


docs/examples/image_features/brief.jl
```diff
@@ -0,0 +1,84 @@
+# ---
+# cover: assets/brief.gif
+# title: BRIEF Descriptor
+# description: This demo shows BRIEF descriptor
+# author: Anchit Navelkar, Ashwani Rathee
```
johnnychen94, line 5: I believe it is `;` here

```suggestion
# author: Anchit Navelkar; Ashwani Rathee
```

johnnychen94:
In the future, we would want to make all these images generated on the fly. And might move back to ImageFeatures own documentation when #159 is finished.

PR #207, opened 2021-07-19
Demo:Local Binary Patterns

ashwani-rathee:
Covers Local Binary Patterns operations
This one is bit difficult to tackle and with lot of content to cover.

Status: Early Stage, Not Ready
Preview Link: https://juliaimages.org/previews/PR207/examples/


johnnychen94 (approved):
I noticed that you have used filename `lbp.png` and `LBP.png`. This could cause confusion because not all file systems are case-sensitive, thus for clarification, I rename them to more distinguishable names.

It is often the case that 512x512 images are too large for demo purposes; it occupies more spaces and thus requires more bandwidth for users to open it up. Thus I used `restrict` here to load a 256x256 image.

PR #209, opened 2021-07-28
Revert "CompatHelper: bump compat for "Documenter" to "0.27""

johnnychen94:
Reverts JuliaImages/juliaimages.github.io#198

Looks like we're having some CI build issues with Documenter v0.27 so let's see if reverting this commit brings the green check back.

cc: @ashwani-rathee


johnnychen94:
Okay, so something upstream breaks the workflow. It might be Plots or something as heavy as it.

Since PR #198 is innocent, I'm closing this one.

PR #210, opened 2021-07-28
Temporarily remove Plots dependency and histogram demo

johnnychen94:
Another try to debug our CI issue


johnnychen94:
@ashwani-rathee so this is clearly a Plots issue.

johnnychen94:
So we decide to merge this PR and bring CI back to the track. When there's new Plots release, we can try and see if it works again.

PR #215, opened 2021-08-10
Demo: GLCM matrix

ashwani-rathee:
- [x] Covers  GLCM(Gray Level Co-occurence Matrix) of an image


docs/examples/image_features/glcm.jl
```diff
@@ -0,0 +1,139 @@
...
+# As we can already take a guess, patch 1 and patch 2 are very similiar(unidirectional texture) and 
+# that's also true for patch 3 and patch 4 which are very similiar(smooth texture).
+
+glcm_results = [];
+glcm_sym_results = [];
+glcm_norm_results = [];
+
+# The `distances` and `angles` arguments may be a single integer or a vector of 
+# integers if multiple GLCMs need to be calculated. The `mat_size` argument is used
+# to define the granularity of the GLCM.
```
kimikage, line 78: It would be nice to have a more specific explanation of what the numbers in `angles` mean.
From the keyword "integer", I imagined those numbers to be discretized values in units of 45°, but looking at the source code, they seem to be in radians. :thinking:

kimikage (commented):
Thank you for your help in improving the documentation.
I don't know much about GLCM, but I just happened to see this, so I leave some trivial comments.

docs/examples/image_features/glcm.jl
```diff
@@ -0,0 +1,139 @@
+# ---
+# cover: assets/glcm.gif
+# title: Gray Level Co-occurrence Matrix
+# description: This demo shows GLCM(Gray Level Co-occurrence Matrix)
```
johnnychen94, line 4: Maybe

```suggestion
# description: The first introduction to GLCM(Gray Level Co-occurrence Matrix)
```

docs/examples/image_features/glcm.jl
```diff
@@ -0,0 +1,139 @@
...
+# author: Ashwani Rathee
+# date: 2021-08-9
+# ---
+
+# Gray Level Co-occurrence Matrix (GLCM) is used for texture analysis. 
+# We consider two pixels at a time, called the reference and the neighbour pixel.
+# We define a particular spatial relationship between the reference and neighbour 
+# pixel before calculating the GLCM. For eg, we may define the neighbour to be 1
```
johnnychen94, line 12:
```suggestion
# pixel before calculating the GLCM. For example, we may define the neighbour to be 1
```

docs/examples/image_features/glcm.jl
```diff
@@ -0,0 +1,139 @@
...
+# Once a spatial relationship is defined, we create a GLCM of size (Range of
+# Intensities x Range of Intensities) all initialised to 0. For eg, a 8 bit single
+# channel Image will have a 256x256 GLCM. We then traverse through the image and for
+# every pair of intensities we find for the defined spatial relationship, we increment
+# that cell of the matrix.
+
+# ## Gray Level Co-occurence Matrix
+
+# Each entry of the GLCM[i,j] holds the count of the number of times that pair of 
+# intensities appears in the image with the defined spatial relationship.
```
johnnychen94, line 25: Hmmm, I actually still don't quite understand the contents of `GLCM` here. Maybe add more explanation here, for example, why is `GLCM[1, 3] == 3` in this case? (I'm using 1-based indexing but this figure seems using 0-based indexing).

docs/examples/image_features/glcm.jl
```diff
@@ -0,0 +1,139 @@
...
+img_patch4 = img_src[60:70, 320:330]
+img_patches = [img_patch1, img_patch2, img_patch3, img_patch4]
+mosaicview(img_patches; nrow=1, npad=1, fillvalue=1)
+
+# As we can already take a guess, patch 1 and patch 2 are very similiar(unidirectional texture) and 
+# that's also true for patch 3 and patch 4 which are very similiar(smooth texture).
+
+glcm_results = [];
+glcm_sym_results = [];
+glcm_norm_results = [];
```
johnnychen94, line 74: Since it's mostly equivalent, I feel we can just introduce one version, say, `glcm`, and leave the other versions to users themselves.

docs/examples/image_features/glcm.jl
johnnychen94: Maybe also add this and explain the arguments to `glcm`.

```julia
X = [0 1 0 2;
     0 2 1 1;
     3 1 0 0;
     0 0 2 3]
glcm(X, 1, 0, 4)
```

docs/examples/image_features/glcm.jl
```diff
@@ -0,0 +1,139 @@
...
+for i in glcm_results
+    point = []
+    for j in property
+        glcm_pro = glcm_prop(i, j)
+        push!(point,glcm_pro)
+    end
+    push!(x,point[1])
+    push!(y,point[2])
+end
+x,y
```
johnnychen94, line 123: I think this can be written in more compact and more efficient form:

```julia
glcm_correlation = [glcm_prop(x, correlation) for x in glcm_results]
glcm_dissimilarity = [glcm_prop(x, dissimilarity) for x in glcm_results]
```

We should try to avoid `push!` as it tries to mutate the underlying memory layout, and doing that can be inefficient.

docs/examples/image_features/glcm.jl
```diff
@@ -0,0 +1,139 @@
...
+
+# These properties can be directly calculated too using syntax `property(glcm_matrix)`.
+# For example: To calculate correlation, we can do `correlation(glcm(img_patch1))``
+
+# We can create graph between correlation and dissimilarity properties of particular
+# GLCM matrices. It's easy to notice that the Patch 1 & Patch 2 are closer in the properties
+# and similiarly for Patch 3 and Patch 4.
+
+# ![](assets/scatter.png)
```
johnnychen94, line 132: Just curious, is it possible to replace the marker dots with the corresponding patch(image)? That would be the most self-explained way I guess.

johnnychen94 (commented):
This is a nice tutorial, and the similarity plot is quite cool.

Here's the structure I have in mind:

- A background introduction to GLCM matrix with some basic math computation on one tiny array (like the figure shows)
- Some introduction to `ImageFeatures.glcm`; how do we link the math concepts to the arguments. For example, `distance`+`angle` seems to define the neighbourhood relationship, and `maxsize` defines the matrix size. And how do we calculate the glcm properties.
- Some applications on `glcm`, in this demo, you have shown the similarity plot using GLCM, which is pretty cool. And then maybe leave some notes or references for readers that are interested in this topic.

How do you think of this tutorial structure?

docs/examples/image_features/glcm.jl
ashwani-rathee: It can be done in Makie.jl but makie keeps crasing in my system for now

PR #216, opened 2021-08-14
Disable notebook generation

johnnychen94:
I find myself never check the generated notebook, so I decide to disable it as we're having quite a long documentation CI.

@ashwani-rathee how do you think of this change? Do you check the notebook often?

---

This requires the latest DemoCards (https://github.com/johnnychen94/DemoCards.jl/pull/98) to actually work, so no need to merge now.

### DIFF
--- a/docs/examples/config.json
+++ b/docs/examples/config.json
@@ -1,3 +1,6 @@
 {
-    "theme": "grid"
+    "theme": "grid",
+    "properties":{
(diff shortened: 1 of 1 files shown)


johnnychen94:
The problem with the generated notebooks is that the binder service takes sooooo long to start a docker, and people just get impatient about this. So let's merge this to cut off 1/3 demo build time

PR #218, opened 2021-08-15
WIP: add remote demos from other packages

johnnychen94:
This is a parallel work to #159 and focuses only on demos.

This setup would grab all the demos from other packages and build a centralized one, for test purposes, I've added two packages:

- https://juliaimages.org/ImageTransformations.jl/dev/examples/
- https://juliaimages.org/ImageFiltering.jl/dev/demos/

Currently this is a playground for https://github.com/johnnychen94/DemoCards.jl/pull/89, but you can check the previews at https://juliaimages.org/previews/PR???


johnnychen94:
Yeah it works:



Eventually, I'd like to make it

```json
{
    "theme": "grid",
    "remote": {
        "transformations": "https://github.com/JuliaImages/ImageTransformations.jl.git:docs/examples"
    }
}
```

or

```json
{
    "theme": "grid",
    "remote": {
        "transformations": {
            "url": "https://github.com/JuliaImages/ImageTransformations.jl.git",
            "subdir": "docs/examples"
        }
    }
}
```

so that we don't need to introduce the external `git submodule` commands.

PR #228, opened 2021-09-05
WIP: Region Growing Algorithm: unseeded and seeded

ashwani-rathee:
- [x] Adds region growing demonstration : both seeded and unseeded
Status: `Initial`


docs/examples/image_segmentation/region_growing.jl
```diff
@@ -0,0 +1,120 @@
...
+# ## Seeded Region Growing
+
+# Seeded region growing segments an image with respect to some user-defined seeds.
+# Each seed is a `(position, label)`` tuple, where `position` is a `CartesianIndex` and
+# label is a positive integer. Each label corresponds to a unique partition of 
+# the image. The algorithm tries to assign these labels to each of the remaining 
+# points. If more than one point has the same label then they will be contribute
+# to the same segment.
+
+img = load("assets/worm.jpg")
```
johnnychen94, line 42: Generally, we should not expect users have an "assets" folder at hand. We should either add it to TestImages or using some explicit `download` tricks.

docs/examples/image_segmentation/region_growing.jl
```diff
@@ -0,0 +1,120 @@
...
+# all the pixels have been assigned to some region.
+
+# `unseeded_region_growing` requires the image `img` and `threshold` as its parameters.
+
+img_src = load("assets/worm.jpg")
+
+# Now, let's segment this image using unseeded_region_growing
+
+seg_small = unseeded_region_growing(img_small, 0.05) # here 0.05 is the threshold
+result = map(i->get_random_color(i), labels_map(seg_small)) .* (1 .-Gray.(img_small))
```
johnnychen94, line 109: Use `IndirectArray` instead. https://github.com/JuliaImages/juliaimages.github.io/blob/a064150e7b8567cd52ea18b59892e7540babece2/docs/examples/image_segmentation/watershed.jl#L80-L86

timholy: Really we should decide about https://github.com/JuliaImages/ImageSegmentation.jl/pull/76 first

docs/examples/image_segmentation/region_growing.jl
ashwani-rathee: I think I'll add worm.png to imgur then

PR #235, opened 2021-09-19
[WIP]: Mean shift algorithm

ashwani-rathee:
- [x] covers mean shift image segmentation algorithm


docs/examples/image_segmentation/mean_shift.jl
```diff
@@ -0,0 +1,89 @@
...
+
+# Mean Shift segmentation is a local homogenization technique that is very useful
+# for damping shading or tonality differences in localized objects.
+# It replaces each pixel with the mean of the pixels in a range-r neighborhood 
+# and whose value is within a distance d.
+
+# K-means segmentation is another segmentation technique which is popular
+# but has two main issues:
+# -  We have specify a prior on the number of clusters(Number of K which is no of
+#    segments).
```
timholy, line 17:
```suggestion
# -  We have specify a prior on the number `K` of clusters (segments).
```

docs/examples/image_segmentation/mean_shift.jl
```diff
@@ -0,0 +1,89 @@
...
+# - It can be sensitive to value of k-means parameter.
```
timholy, line 18: This seems not entirely surprising. Is it really worth making as a separate point?

It could also be stated that K-means tends to return clusters of approximately equal radius, which may not be how a human would do it.

docs/examples/image_segmentation/mean_shift.jl
```diff
@@ -0,0 +1,89 @@
...
+
+# Below example image shows x,y coordinates corresponds to features and 3rd column 
+# is the density of the feature.
+
+# ![](assets/normalized.png)
+
+# - Each hill represents a cluster
+# - Peak(mode) of hill represents "center" of the cluster
+# - Each pixel climbs the steepest hill within its neighborhood
+# - Pixel assigned to the hill(cluster) it climbs. That's the idea behind meanshift.
```
timholy, line 34:
```suggestion
# - A pixel is assigned to the hill(cluster) it climbs.
```
I think the "idea behind meanshift" encompasses all of these points.

docs/examples/image_segmentation/mean_shift.jl
```diff
@@ -0,0 +1,89 @@
...
+# ![](assets/coloredpfd.png)
+
+# The Mean Shift takes usually 3 inputs:
+
+# - A distance function for measuring distances between pixels. Usually the 
+#   Euclidean distance, but any other well defined distance function could be used. 
+#   The Manhattan Distance is another useful choice sometimes.
+# - A radius. All pixels within this radius (measured according the above distance) 
+#   will be accounted for the calculation.
+# - A value difference. From all pixels inside radius r, we will take only those
```
timholy, line 46: Would "intensity" be clearer than "value"?

docs/examples/image_segmentation/mean_shift.jl
```diff
@@ -0,0 +1,89 @@
...
+
+# Given: Distribution of N pixels in feature space.
+
+# Task: Find modes(clusters) of distribution. 
+
+# Clustering:
+# - Set $m_i = f_i$ as initial mean fro each pixel idea
+# - Repeat the following for each mean mi:
+#   - place window of size $w$ around $m_i$.
+#   - compute centroid m within the window set $mi = main$
```
timholy, line 59: Not sure what `main` means

docs/examples/image_segmentation/mean_shift.jl
```diff
@@ -0,0 +1,89 @@
...
+#   whose values are within this difference for calculating the mean
+
+# ## Mean Shift Algorithm
+
+# Given: Distribution of N pixels in feature space.
+
+# Task: Find modes(clusters) of distribution. 
+
+# Clustering:
+# - Set $m_i = f_i$ as initial mean fro each pixel idea
```
timholy, line 56:
```suggestion
# - Set $m_i = f_i$ as initial mean for each pixel
```

docs/examples/image_segmentation/mean_shift.jl
```diff
@@ -0,0 +1,89 @@
...
+# - Repeat the following for each mean mi:
```
timholy, line 57: Use the same notation. Is `mi` the same as `$m_i$`?

docs/examples/image_segmentation/mean_shift.jl
```diff
@@ -0,0 +1,89 @@
...
+#   - place window of size $w$ around $m_i$.
+#   - compute centroid m within the window set $mi = main$
+#   - stop if shift in mean $m_i$ is less than a threshoold $\epsilon$, $m_i$ is the mode
+
+# It's a very simple but computationally expensive method which doesn't require input of
+# the number of clusters. Very robust to outliers and clustering depends on window size w.
```
timholy, line 63: Technically, one could say that it replaces specification of # of clusters with the size of the radius.

timholy (commented):
I hadn't noticed these (they arrived back when I wasn't paying attention), thanks for adding this!

PR #240, opened 2021-11-04
advertise a few packages of Peter Kovesi

johnnychen94:
@peterkovesi feel free to make suggestions if you want to rephrase the words. I didn't add https://github.com/peterkovesi/PerceptualColourMaps.jl here as it's perhaps related to Colors.jl more. We can add it there if you want.


peterkovesi:
Thanks for the mentions of my packages.  PerceptualColourMaps is more relevant to the plotting packages and it gets a mention in a number of plotting example pages. So that is fine

PR #248, opened 2022-06-28
acknowledge tongyuan's support

johnnychen94:
As some of you already know, I work part-time for Tongyuan since this April to build their MATLAB-flavor image processing toolbox. They generously allow me to upstream the major part of the codes to JuliaImages so I figure I should mention the name somewhere in JuliaImages.


timholy (approved):
I suppose I should also add some of my own grant funding?

